PR #11, opened 2025-04-17
feat(trackers): add support for pytorch rocm

### LINKED ISSUE #10: Pytorch OpenCL
Can it theoretically work not only with pytorch cuda but also with pytorch opencl backend?

soumik12345:
# Description

Adds support for pytorch rocm backend using the extra `rocm`.

Fixes #10 

## Type of change

Please delete options that are not relevant.

-   [ ] This change requires a documentation update


.github/workflows/test-doc.yml
```diff
@@ -30,8 +30,7 @@ jobs:
 
       - name: 🏗️ Install dependencies
         run: |
-          uv pip install -r pyproject.toml --group docs
-
+          uv pip install ".[docs,cpu]"
```
SkalskiP: @onuralpszr wdyt?

pyproject.toml
```diff
@@ -89,17 +93,20 @@ conflicts = [
     [
         { extra = "cpu" },
         { extra = "cu124" },
```
SkalskiP, line 117: wouldn't we want to add support for more environments?

SkalskiP:
@soumik12345 let's use this opportunity to move `validators` to code dependencies.

.github/workflows/test-doc.yml
onuralpszr: docs not exist so answer is no it should "uv pip install -r pyproject.toml --group docs --extra cpu" 

in pypi we can only use "cpu", as I said I excluded "docs"

onuralpszr:
@SkalskiP  @soumik12345  I am squashing some of the commit spams for cleaner look I already got permission from @soumik12345  so all clear.  Also adding extra tests and found a bug

pyproject.toml
```diff
@@ -37,22 +37,29 @@ dependencies = [
...
+
 cu124 = [
     "torch>=2.6.0",
     "torchvision>=0.21.0",
 ]
 
+rocm61 = [
+    "torch",
+    "torchvision",
+]
```
soumik12345: @onuralpszr Should we pin the torch versions for rocm?

trackers/__init__.py
```diff
@@ -1,5 +1,14 @@
...
+try:
+    from trackers.core.deepsort.feature_extractor import DeepSORTFeatureExtractor
+    from trackers.core.deepsort.tracker import DeepSORTTracker
+
+    __all__.extend(["DeepSORTFeatureExtractor", "DeepSORTTracker"])
+except ImportError:
+    print(
+        "DeepSORT dependencies not installed. DeepSORT features will not be available."
+    )
+    pass
```
soumik12345: Nice!

.github/workflows/test-doc.yml
SkalskiP: @soumik12345 I'm tagging you to make sure you'll see this comment from @onuralpszr

soumik12345: @onuralpszr Does this mean that we won't be able to use the cuda and rocm extras from pypi?

trackers/__init__.py
SkalskiP: @soumik12345 this is great, but I'd also want to include a recommendation on what to do to fix it. Here's [how](https://github.com/roboflow/rf-detr/blob/a9851a3f5fe611437663393453438ff1cc28c6a1/rfdetr/main.py#L482) we did it in RF-DETR.

.github/workflows/test-doc.yml
SkalskiP: very good question; @onuralpszr how will this influence regular `pip` users of the final package? will they be able to install `cpu`, `cuda`, ... ?

if not, is there any way we can unlock them?

pyproject.toml
onuralpszr: @soumik12345 I am going to check min py3.9 for all and others then we can pin "min" for all

.github/workflows/test-doc.yml
onuralpszr: @SkalskiP  @soumik12345  all extras we made like "cpu", "cuda" all fine I only exclude "docs,devs,build" "--group" means for local development as "uv" documentation stated. For rest we are ALL good :)

trackers/__init__.py
soumik12345: Yeah, IMO including a pip installation instruction would be helpful. We also need to author a strong installation instructions document.

soumik12345: Probably we should also have installation instruction for every tracker in their respective docs.

onuralpszr: I included pip install method to except section 

```python
from trackers.core.sort.tracker import SORTTracker

__all__ = ["SORTTracker"]

try:
    from trackers.core.deepsort.feature_extractor import DeepSORTFeatureExtractor
    from trackers.core.deepsort.tracker import DeepSORTTracker

    __all__.extend(["DeepSORTFeatureExtractor", "DeepSORTTracker"])
except ImportError:
    print(
        "DeepSORT dependencies not installed. DeepSORT features will not be available. "
        "Please run `pip install trackers[deepsort]` and try again."
    )
    pass

```

pyproject.toml
onuralpszr: I added min for all.

soumik12345 (commented):
LGTM

PR #17, opened 2025-04-20
feat(trackers): Implementation of Kalman Gating for `DeepSORTTracker`

soumik12345:
# Description

Implementation of Kalman motion gating described in Section 2.2 of the [DeepSORT paper](https://arxiv.org/pdf/1703.07402). This can be used to prune unlikely detection‐to‐track associations based on the predicted uncertainty of each track’s state.


trackers/core/deepsort/tracker.py
```diff
@@ -282,37 +356,92 @@ def _get_associated_indices(
...
+                    for j, det_idx in enumerate(feasible_detections):
+                        if gating_distances[j] > MAHALANOBIS_THRESHOLD:
+                            combined_dist_matrix[tracker_idx, det_idx] = (
+                                1.0  # Mark as infeasible
+                            )
+
+        confirmed_matches, unmatched_confirmed, unmatched_detections = (
+            self._match_tracks_stage(
+                combined_dist_matrix,
+                confirmed_tracks,
```
rolson24, line 290: I think technically the original paper specified that DeepSORT also uses the Mahalanobis distance to combine with the appearance matching distance, not the IOU distance.

See here in section 2.2:
"
In combination, both metrics complement each other by serving different aspects of the assignment problem. On
the one hand, the Mahalanobis distance provides information about possible object locations based on motion that are particularly useful for short-term predictions. On the other hand, the cosine distance considers appearance information that are particularly useful to recover identities after longterm occlusions, when motion is less discriminative. To build the association problem we combine both metrics using a
weighted sum
$$c_{i,j} = λ d^{(1)}(i, j) + (1 − λ)d^{(2)}(i, j)$$ (5)
"

rolson24 (commented):
Mahalanobis distance calculation looks good! I added a comment on the combined cost matrix. Also I was reading through the paper again and noticed an interesting section when they describe the matching cascade in section 2.3 and the pseudocode where they show the algorithm:
"
Therefore, we introduce a matching cascade that gives priority to more frequently seen objects to encode our notion of probability spread in the association likelihood. Listing 1 outlines our matching algorithm. As input we provide the set of track T and detection D indices as well as the maximum age $A_{max}$. In lines 1 and 2 we compute the association cost matrix and the matrix of admissible associations. We then iterate over track age n to solve a linear assignment problem for tracks of increasing age. 
"

Which makes it sound like instead of just strictly sorting by lowest cost, the paper first sorts by time since last match (they call this track age) and then does matching for each set of track ages. So they would first calculate the cost matrix for tracks that were matched in the previous frame and do matching with the detections, then they would move onto the tracks that had been lost for one frame and calculate a new cost matrix with the leftover detections and do matching again and so on. I am not sure if we want to implement this, but I think that is what the paper says.
If we did implement it, I think the best way would be to change `_get_associated_inds` to loop over the different groups of tracks that have the same `time_since_update` and do the distance calculation and matching for each group. In each group of tracks, the tracks that don't get matched are marked lost, and the detections that don't get matched get sent onto the next older group of tracks to be matched.

trackers/utils/sort_utils.py
```diff
@@ -143,3 +143,26 @@ def update_detections_with_track_ids(
     updated_detections.tracker_id = np.array(final_tracker_ids)
 
     return updated_detections
+
+
+def convert_bbox_to_xyah(state_bbox: np.ndarray) -> np.ndarray:
```
SkalskiP: Let’s stick with the naming convention we use in `supervision`. I suggest renaming this function to `xyxy_to_xcycarh`. @soumik12345, could you open a PR to `supervision` and add this function—along with unit tests—to `supervision/detection/utils`?

SkalskiP: Let's rename input argument from `state_bbox` to `xyxy` as well.

onuralpszr: @SkalskiP  I actually asked and we talked with @soumik12345  already so PR in progress : https://github.com/roboflow/supervision/pull/1823  😅

trackers/utils/sort_utils.py
```diff
@@ -143,3 +143,25 @@ def update_detections_with_track_ids(
...
+    where the aspect ratio is `width / height`.
+
+    Args:
+        xyxy (np.ndarray): Bounding box in format `(x1, y1, x2, y2)`.
+
+    Returns:
+        np.ndarray: Bounding box in format
+            `(center x, center y, aspect ratio, height)`.
+    """
+    x1, y1, x2, y2 = xyxy
```
onuralpszr:
```suggestion
    x1, y1, x2, y2 = xyxy.T
```

onuralpszr (changes requested):

SkalskiP:
@soumik12345 As we agreed in Chicago, this PR is on hold until we’re able to measure the quality of the tracker output.

PR #7, opened 2025-04-15
Feat: Evaluation framework

rolson24:
# Description

This is a PR implementing an evaluation framework for evaluating trackers. The framework allows user's to use any detector they want by writing a callback function that takes in an frame and expects a sv.Detections object to be returned. The evaluation framework can currently handle the MOT Challenge dataset format, and implements CLEAR metrics. The framework handles evaluating trackers not implemented in "Trackers" by allowing users to either provide a tracking callback function that takes in a frame and returns an sv.Detections object, or just provide a "Trackers" tracker object. 

List any dependencies that are required for this change.

## Type of change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] This change requires a documentation update

## Example usage:

```python
import numpy as np
import supervision as sv
from rfdetr import RFDETRBase
from rfdetr.util.coco_classes import COCO_CLASSES
from trackers import SORTTracker
from trackers.eval import evaluate_tracker

model = RFDETRBase(device="mps")
tracker = SORTTracker()

mot_dataset_path = "path/to/MOT17/train"
mot_dataset = MOTChallengeDataset(dataset_path=mot_dataset_path)


def detection_callback(frame: np.ndarray, frame_info: str):
    detections = model.predict(frame, threshold=0.5)
    # With a different model, need to convert to sv.Detections
    return detections

results = evaluate_tracker(
    dataset=mot_dataset,
    detection_source=detection_callback  # Could also use mot_dataset to use public detections from MOTChallengeDataset
    tracker_source=tracker,
    metrics=["Count", "CLEAR"],  # Specify desired metrics
    cache_tracks=True,  # Example: cache the tracks
    cache_dir="./cached_tracks_sv",  # Example cache directory
)

print("\n--- Evaluation Results ---")
print(json.dumps(results, indent=2))
```

## Example usage with tracker callback:
```python
import numpy as np
import supervision as sv
from rfdetr import RFDETRBase
from rfdetr.util.coco_classes import COCO_CLASSES
from BoxMOT import ByteTrack
from trackers.eval import evaluate_tracker

model = RFDETRBase(device="mps")

mot_dataset_path = "path/to/MOT17/train"
mot_dataset = MOTChallengeDataset(dataset_path=mot_dataset_path)


def detection_callback(frame: np.ndarray, frame_info: str):
    detections = model.predict(frame, threshold=0.5)
    # With a different model, need to convert to sv.Detections
    return detections

[...]

PR #29, opened 2025-04-26
Feat: MOT dataset loader

rolson24:
# Description

This PR is the start of the evaluation framework started in #7. It contains a base dataset prototype and a MOTChallenge dataset implementation. The MOTChallenge dataset class can load, parse, preprocess, and provide an iterator for each sequence in the dataset, much in the same way [TrackEval](https://github.com/JonathonLuiten/TrackEval/tree/12c8791b303e0a0b50f753af204249e622d0281a) does. It also supports loading "public detections" which are pre-computed detections stored in a `dets.txt` file for each sequence which standardizes the detections that object trackers are evaluated on.

## Type of change

Please delete options that are not relevant.

-   [ ] This change requires a documentation update

Example usage:

```python
from pathlib import Path
import supervision as sv
from trackers.dataset.mot_challenge import MOTChallengeDataset

local_mot17_path = "/path/to/MOT17"
dataset_root = os.path.join(local_mot17_path, 'train')
mot_dataset_path = Path(dataset_root)

# Initialize the dataset with the path to the MOT folder.
mot_dataset = MOTChallengeDataset(dataset_path=mot_dataset_path)

# Print the sequences in the dataset
available_sequences = mot_dataset.get_sequence_names()
print("Available sequences:", available_sequences)

# Print the info of the first sequence
first_sequence_name = available_sequences[0]
seq_info = mot_dataset.get_sequence_info(first_sequence_name)
print("Sequence Info:", seq_info)

# Get frames from the iterator
frame_count = 0
for frame_info in mot_dataset.get_frame_iterator(first_sequence_name):
    if frame_count < 5:
        print(frame_info)
    frame_count += 1
print(f"Total frames iterated: {frame_count}")

# Load the ground truth tracks for the first sequence
gt_detections = mot_dataset.load_ground_truth(first_sequence_name) # sv.Detections object
for i in range(min(5, len(gt_detections))):
    print(f"Frame: {gt_detections.data['frame_idx'][i]}, \
            ID: {gt_detections.tracker_id[i]}, \
            Class: {gt_detections.class_id[i]}, \
            Conf/Vis: {gt_detections.confidence[i]:.2f}, \
            Box: {gt_detections.xyxy[i]}")

# Load public detections for the sequence
mot_dataset.load_public_detections(min_confidence=0.1) # Optional: set a confidence threshold

if mot_dataset.has_public_detections:
    frame_iter = mot_dataset.get_frame_iterator(sequence_name)
    first_frame_info = next(frame_iter)
    # Get the image path for the first frame
    first_frame_path = first_frame_info.get('image_path')

    if first_frame_path:
        print(f"Getting detections for image: {first_frame_path}")
        public_dets_frame1 = mot_dataset.get_public_detections(first_frame_path)
        print(f"Found {len(public_dets_frame1)} public detections for the first frame.")
        if len(public_dets_frame1) > 0:
            print("First 5 public detections for this frame:")
            for i in range(min(5, len(public_dets_frame1))):
                print(f"Conf: {public_dets_frame1.confidence[i]:.2f}, \
                        Box: {public_dets_frame1.xyxy[i]}")

```

I also have a [colab notebook](https://colab.research.google.com/drive/1LPoFlgXxf1mGbAFyv13SB_bngLPM1E5g?usp=sharing) testing out the data loader

For example, documentation changes, usability, usage/costs, secrets, etc.

## Docs

-   [ ] Docs updated? What were the changes:


trackers/dataset/mot_challenge.py
```diff
@@ -0,0 +1,602 @@
...
+            logger.warning("No public detections were loaded.")
+
+    @property
+    def has_public_detections(self) -> bool:
+        """Returns True if public detections have been loaded via
+        `load_public_detections`."""
+        return self._public_detections is not None
+
+    def get_public_detections(self, image_path: str) -> sv.Detections:
+        """
```
rolson24, line 434: The keys for the public detections right now are the absolute image paths. Maybe we should change this to frame index and sequence name? Maybe it should return an iterator of detections? This function allows us to use the public detections in a similar way to passing a detector a frame and it returning a set of detections.

soumik12345: I think the best way to tackle this is to split this into 2 functions: `get_public_detections_from_image_path` and `get_public_detections_from_from_index`.

trackers/dataset/base.py
```diff
@@ -0,0 +1,87 @@
+import abc
+from typing import Any, Dict, Iterator, List, Optional, Tuple
+
+import supervision as sv
+
+
+# --- Base Dataset ---
+class Dataset(abc.ABC):
```
soumik12345: Since we will also be working on a dataset class for our Re-ID training pipeline, let's call this `EvaluationDataset` to avoid confusion.

trackers/dataset/utils.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+import supervision as sv
+
+from trackers.log import get_logger  # Added import
+
+logger = get_logger(__name__)  # Added logger instance
+
+
+def _relabel_ids(detections: sv.Detections) -> sv.Detections:
```
soumik12345: Any reason for this to be a private function?

trackers/dataset/utils.py
```diff
@@ -0,0 +1,86 @@
...
+            "Could not convert tracker IDs to integers during relabeling. Skipping."
+        )
+        return detections
+
+    if len(unique_ids) == 0:
+        return detections
+
+    # Now unique_ids contains only valid integers
+    max_id = np.max(unique_ids)
+    min_id = np.min(unique_ids)
```
soumik12345: Adding type annotations for these 2 variables should fix the pre-commit ci.

trackers/dataset/base.py
```diff
@@ -0,0 +1,87 @@
...
+    """Abstract base class for datasets used in tracking evaluation."""
+
+    @abc.abstractmethod
+    def load_ground_truth(self, sequence_name: str) -> Optional[sv.Detections]:
```
soumik12345, line 12: It seems like `Union[sv.Detections, None]` would be a more suitable return type here. @onuralpszr WDYT?

soumik12345 (changes requested):
Thanks a lot for this PR!
Kudos for attaching such a detailed colab notebook with the PR, really helped me test the code quickly!

I have recommended some minor changes. Can you please address them?

trackers/dataset/base.py
onuralpszr: Optional[...] is a shorthand notation for Union[..., None] so it is the same thing :)

trackers/dataset/mot_challenge.py
rolson24: That makes sense

trackers/dataset/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    if len(detections) == 0 or detections.tracker_id is None:
+        return detections
+
+    # 1. Filter out potential NaN values first
+    valid_ids_mask = ~np.isnan(detections.tracker_id)
+    if not np.any(valid_ids_mask):
+        # All IDs were NaN or array was empty after filtering
+        return detections
+
+    # 2. Get unique integer IDs
```
soumik12345: Same as the previous comment.

trackers/dataset/utils.py
```diff
@@ -0,0 +1,86 @@
...
+            need relabeling.
+
+    Returns:
+        sv.Detections: The detections object with relabeled `tracker_id`s.
+        Returns the original object if no valid IDs are found or if input is empty.
+    """
+    if len(detections) == 0 or detections.tracker_id is None:
+        return detections
+
+    # 1. Filter out potential NaN values first
```
soumik12345: Probably it's best to avoid numbered steps in comments. This is because if the logic gets changed, there's a chance that the numbering would need to change as well.

soumik12345 (dismissed):
Hi @rolson24, the PR looks good to me.
I have some minor nitpick comments.

trackers/dataset/utils.py
rolson24: Fixed nits

trackers/dataset/base.py
```diff
@@ -0,0 +1,87 @@
+import abc
+from typing import Any, Dict, Iterator, List, Optional, Tuple
+
+import supervision as sv
+
+
+# --- Base Dataset ---
```
SkalskiP, line 7: let’s remove Python comments like these; we can already tell it’s a base dataset since it only has abstract methods and implements the ABC interface.

trackers/dataset/base.py
```diff
@@ -0,0 +1,87 @@
...
+    def get_sequence_info(self, sequence_name: str) -> Dict[str, Any]:
+        """
+        Returns metadata for a specific sequence.
+
+        Args:
+            sequence_name: The name of the sequence.
+
+        Returns:
+            A dictionary containing sequence information (e.g., 'frame_rate',
+            'seqLength', 'img_width', 'img_height', 'img_dir'). Keys and value
```
SkalskiP, line 41: can we make sure all these dict keys follow the same naming convention? It looks like everything is in snake_case except for `seqLength`. Also, wouldn’t it make more sense to use a `dataclass` or a `namedtuple` instead of a regular dict here?

trackers/dataset/base.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        pass
+
+    @abc.abstractmethod
+    def preprocess(
+        self,
+        ground_truth: sv.Detections,
+        predictions: sv.Detections,
+        iou_threshold: float = 0.5,
+        remove_distractor_matches: bool = True,
```
SkalskiP, line 67: it seems that the preprocess implementation is highly dataset-specific. given that, does it really make sense to add `iou_threshold` and `remove_distractor_matches` to the base class, especially since there’s no guarantee they’ll be needed here? maybe it would be better to allow for additional keyword arguments instead?

trackers/dataset/base.py
```diff
@@ -0,0 +1,87 @@
+import abc
+from typing import Any, Dict, Iterator, List, Optional, Tuple
+
+import supervision as sv
+
+
+# --- Base Dataset ---
+class EvaluationDataset(abc.ABC):
```
SkalskiP, line 8: should we really call this dataset `EvaluationDataset`? Maybe it would make more sense to use a more general name, like `TrackingDataset` or `MOTDataset` instead?

trackers/dataset/utils.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+import supervision as sv
+
+from trackers.log import get_logger
+
+logger = get_logger(__name__)
+
+
+def relabel_ids(detections: sv.Detections) -> sv.Detections:
```
SkalskiP, line 9: - Let’s rename this function from `relabel_ids` to `remap_ids`.
- We should also allow users to pass a custom mapping that defines how each `source_id` should be mapped to a `target_id`.
- I definitely think we should add a unit test for this function.

trackers/dataset/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    try:
+        unique_ids = np.unique(detections.tracker_id[valid_ids_mask].astype(int))
+    except ValueError:
+        logger.warning(
+            "Could not convert tracker IDs to integers during relabeling. Skipping."
+        )
+        return detections
+
+    if len(unique_ids) == 0:
+        return detections
```
SkalskiP, line 43: do we need this?

trackers/dataset/mot_challenge.py
```diff
@@ -0,0 +1,635 @@
...
+        frame_detections: Dict[int, List[Dict[str, Any]]] = {}
+
+        try:
+            with open(file_path, "r") as f:
+                for line in f:
+                    line = line.strip()
+                    if not line:
+                        continue
+
+                    parts = line.split(",")
```
SkalskiP, line 115: `supervision` have a [util](https://github.com/roboflow/supervision/blob/95f2b54cc8c8dce16658c95cd3376953d1cd1670/supervision/utils/file.py#L60) for loading text file lines; let's use it here

trackers/dataset/mot_challenge.py
```diff
@@ -0,0 +1,635 @@
...
+                        y = float(parts[3])
+                        width = float(parts[4])
+                        height = float(parts[5])
+                        # --- Ensure 7th column (index 6) is assigned to confidence ---
+                        # This value often represents visibility/ignore flag in gt.txt
+                        # (0=ignore)
+                        # or detection confidence in det.txt
+                        confidence = float(parts[6])
+
+                        class_id = int(parts[7]) if len(parts) > 7 else -1
```
SkalskiP, line 135: I'd move `-1` magic number to the top of the file and set is as python constant.

trackers/dataset/mot_challenge.py
```diff
@@ -0,0 +1,635 @@
...
+            frame_idx = np.array(frame_indices, dtype=np.int32)
+
+            # Create sv.Detections object with frame indices in data
+            return sv.Detections(
+                xyxy=xyxy,
+                confidence=confidence,
+                class_id=class_id,
+                tracker_id=tracker_id,
+                data={"frame_idx": frame_idx},
+            )
```
SkalskiP, line 218: I think this method must return separate `Detections` object per frame.

trackers/dataset/mot_challenge.py
```diff
@@ -0,0 +1,635 @@
...
+        """
+        sequences = []
+        for item in self.root_path.iterdir():
+            if item.is_dir() and (item / "seqinfo.ini").exists():
+                sequences.append(item.name)
+        if not sequences:
+            logger.warning(f"No valid MOTChallenge sequences found in {self.root_path}")
+        return sorted(sequences)
+
+    def _parse_mot_file(
```
SkalskiP, line 80: I think we would need to separate loading data from file and processing data; this will allow us to unit test the data processing part

trackers/dataset/mot_challenge.py
```diff
@@ -0,0 +1,635 @@
...
+        if not potential_files:
+            logger.warning(
+                f"No image files found matching pattern '{first_frame_pattern}' \
+                    in {img_dir}"
+            )
+            # Try common extensions explicitly if glob fails
+            if (img_dir / f"{1:06d}.jpg").exists():
+                img_ext = ".jpg"
+            elif (img_dir / f"{1:06d}.png").exists():
+                img_ext = ".png"
```
SkalskiP, line 326: I'm a bit confused as to how any of this would work if `glob` fails.

SkalskiP (changes requested):

SkalskiP:
Hi @rolson24 👋🏻 After my conversation with @soumik12345 yesterday, it became clear to me that we really need to enable benchmarking in the trackers package as soon as possible. Because of this, getting this PR over the finish line is my top priority for the week. I’ve left my comments on the PR—do you think you’ll have time to address them? If not, no worries, I can take care of it to keep things moving forward.

rolson24:
> Hi @rolson24 👋🏻 After my conversation with @soumik12345 yesterday, it became clear to me that we really need to enable benchmarking in the trackers package as soon as possible. Because of this, getting this PR over the finish line is my top priority for the week. I’ve left my comments on the PR—do you think you’ll have time to address them? If not, no worries, I can take care of it to keep things moving forward.

Hi @SkalskiP, I can address these request these changes this weekend. Sorry I was finishing up school last week, but I should have some more time now.

trackers/dataset/mot_challenge.py
rolson24: So instead of storing the frame index in the `data` attribute of the `Detections` object, we should just return a list of `Detections` objects? That probably does align more with how supervision uses `Detections` objects.

SkalskiP:
Hi @rolson24 👋🏻 Thanks a lot for all the work you did for trackers. After long break we are back working on this project. This feature is now implemented and it's going to be released in `trackers-2.2.0`.

PR #32, opened 2025-04-29
feat(trackers): Implement linear sum assignment for `SORTTracker`

soumik12345:
# Description

Implement linear sum assignment to associate detections with trackers in the `SORTTracker`.


AlexBodner (commented):
LGTM

PR #33, opened 2025-04-29
fix(trackers): initializing feature extractor from timm

soumik12345:
# Description

Fix feature extractor initialization from timm.


AlexBodner (approved):
This change makes the feature extraction class able to convert the model to the feature extractor. Tried it with ByteTrack and also works as expected.

PR #34, opened 2025-04-30
feat(trackers): Update the LSA computation for DeepSORT

soumik12345:
Update the linear sum assignment implementation for DeepSORT.


rolson24 (approved):
Look good!

PR #35, opened 2025-04-30
Update from_timm from feature_extractor of DeepSORT

AlexBodner:
# Description

The argumented pretrained was not read when it was passed by **kwargs, I had to specify pretrained = pretrained in order to get a pretrained model.

List any dependencies that are required for this change.

By printing the distance matrix between features, it has some meaningful values only when calling it like I propose. I invite you to print out the appearance distance matrix in _get_associated_indices


soumik12345 (approved):
LGTM!

PR #37, opened 2025-05-01
add: Market1501SiameseDataset

soumik12345:
# Description

An implementation of the Marker-1501 dataset for training a siamese network using triplet loss.


trackers/utils/data_utils.py
```diff
@@ -0,0 +1,22 @@
+import zipfile
+
+
+def unzip_file(zip_filepath, extract_to_path):
```
SkalskiP: - Let's include input and output types.

- In `supervision`, we follow the `source` and `target` prefix naming convention. I'd like to keep this consistent in the trackers package as well. I suggest renaming `zip_filepath` and `extract_to_path` to `source_zip_path` and `target_dir_path`.

tests/test_reid_dataset.py
```diff
@@ -0,0 +1,22 @@
+import os
```
SkalskiP: As I mentioned before, I'm not a fan of unit tests that create a false sense of security. There are quite a few aspects of the `Market1501Dataset` that we’re not testing, yet it might give the misleading impression that the component is fully covered. I think we can skip adding unit tests for now. However, if we do want to include them, we should cover much more than just the length.

trackers/core/reid/data/base.py
```diff
@@ -0,0 +1,58 @@
...
+        self.data_dir = data_dir
+        self.transforms = transforms or ToTensor()
+        self.triplet_classes = self.get_triplet_classes()
+
+    @abstractmethod
+    def get_triplet_classes(self):
+        pass
+
+    @abstractmethod
+    def get_anchor_image_file(self, triplet_class: str) -> str:
```
SkalskiP: I don’t think `TripletsDataset` should require implementing `get_anchor_image_file`, `get_positive_image_file`, and `get_negative_image_file`. Instead, it should expect a standardized `Dict[str, List[str]]` mapping `tracker_id` to a list of image paths. Then the dataset’s responsibility would be just selecting and loading triplets. This dictionary could either be returned by an abstract method (implemented by a child class) or passed directly via the constructor. As you know, I’m not a big fan of inheritance, so I’d lean toward the second option.

trackers/core/reid/data/market_1501.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    def __init__(self, data_dir: str, transforms: Optional[Compose] = None):
+        super().__init__(data_dir=data_dir, transforms=transforms)
+        self.train_data_dir = os.path.join(data_dir, "bounding_box_train")
+        self.secure_random = SystemRandom()
+
+    def get_triplet_classes(self):
+        train_data_dir = os.path.join(self.data_dir, "bounding_box_train")
+        image_files = glob(os.path.join(train_data_dir, "*.jpg"))
```
SkalskiP: With the current architecture, we're running `glob` three times for every triplet. As I mentioned in my previous comment, `TripletsDataset` should work with a pre-parsed, standardized `Dict[str, List[str]]`, where keys represent classes and values are lists of image paths associated with each class. By the way, I'm not sure if "classes" is the most accurate term here—if I understand correctly, these could be either `class_id`s or `tracker_id`s.

SkalskiP:
Hi @soumik12345

I spent some time today reviewing your PR, taking https://github.com/roboflow/trackers/pull/29 as context. Below you’ll find a base implementation of `TripletsDataset` and `Market1501Dataset`, along with comments that outline my thought process.

- `TripletsDataset` should expect a standardized `Dict[str, List[str]]` dictionary mapping `tracker_id` / `class_id` (we might come up with a better, more general name) to a list of image paths representing that object. It should handle all logic related to selecting and loading triplets internally.
- Child classes like `Market1501Dataset` should focus solely on parsing the dataset structure into the `Dict[str, List[str]]` format expected by `TripletsDataset`.
- It should be easy to define a method for converting `MOTChallengeDataset` (https://github.com/roboflow/trackers/pull/29) into a `Dict[str, List[str]]` that we can then inject into `TripletsDataset`.
- The implementation of `TripletsDataset` below is a baseline. We'll still need to add random seed control and validation for the input dictionary (e.g., ensuring we have at least two distinct `tracker_id` / `class_id`, and filtering out entries with fewer than two images).
- `TripletsDataset` is now testable without having to download datasets or mock file structures on disk.
- If we want to test the correctness of the Market1501 parsing logic, we can do so by testing only `parse_market1501_dataset`. By mocking the output of `glob`, we can avoid touching the file system altogether.

```python
from typing import Dict, List, Optional, Tuple
import os
from glob import glob
import random

import torch
from PIL import Image
from torch.utils.data import Dataset
from torchvision.transforms import Compose, ToTensor


class TripletsDataset(Dataset):
    def __init__(
        self, 
        tracker_id_to_images: Dict[str, List[str]], 
        transforms: Optional[Compose] = None
    ):
        self.tracker_id_to_images = tracker_id_to_images
        self.transforms = transforms or ToTensor()
        self.tracker_ids = list(tracker_id_to_images.keys())
        
    def __len__(self) -> int:
        return len(self.valid_tracker_ids)

    def __getitem__(
        self, index: int
    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
        anchor_tracker_id = self.valid_tracker_ids[index]
        
        anchor_images = self.tracker_id_to_images[anchor_tracker_id]
        anchor_image_path, positive_image_path = random.sample(anchor_images, 2)
        
        negative_tracker_id = random.choice([
            tracker_id 
            for tracker_id 
            in self.valid_tracker_ids 
            if tracker_id != anchor_tracker_id
        ])

        negative_image_path = random.choice(self.tracker_id_to_images[negative_tracker_id])
        
        anchor_image = self._load_and_transform_image(anchor_image_path)
        positive_image = self._load_and_transform_image(positive_image_path)
        negative_image = self._load_and_transform_image(negative_image_path)
        
        return anchor_image, positive_image, negative_image
    
    def _load_and_transform_image(self, image_path: str) -> torch.Tensor:
        image = Image.open(image_path).convert("RGB")
        if self.transforms:
            image = self.transforms(image)
        return image


def parse_market1501_dataset(data_dir: str) -> Dict[str, List[str]]:
    train_data_dir = os.path.join(data_dir, "bounding_box_train")
    image_files = glob(os.path.join(train_data_dir, "*.jpg"))
    
    tracker_id_to_images = {}
    for image_path in image_files:
        tracker_id = os.path.basename(image_path).split('_')[0]
        
        if tracker_id not in tracker_id_to_images:
            tracker_id_to_images[tracker_id] = []
            
        tracker_id_to_images[tracker_id].append(image_path)
        
    return tracker_id_to_images


class Market1501Dataset(TripletsDataset):
    def __init__(self, data_dir: str, transforms: Optional[Compose] = None):
        tracker_id_to_images = parse_market1501_dataset(data_dir)
        super().__init__(tracker_id_to_images=tracker_id_to_images, transforms=transforms)
```

SkalskiP (changes requested):

tests/test_reid_dataset.py
soumik12345: I modified the testcase; it not only tests the length of the datasets but also checks the validity of triplet paths for each unique ID.

trackers/core/reid/data/base.py
```diff
@@ -0,0 +1,78 @@
+import secrets
```
SkalskiP: Let's use `random` instead of `secrets`. I know `secrets` is more secure, but here we're just picking images for a triplet—not generating RSA keys. `random` will do just fine for this use case.

If `bandit` complains, you should be able to silence the specific warning using `# nosec` at the end of the line.

trackers/utils/data_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    if k == 1:
+        return [secrets.choice(population)]
+
+    # For multiple samples, shuffle and take first k
+    result = list(population)
+    for i in range(len(result) - 1, 0, -1):
+        j = secrets.randbelow(i + 1)
+        result[i], result[j] = result[j], result[i]
+    return result[:k]
```
SkalskiP: Once we switch to `random`, we can use `random.sample` and remove this custom function.

trackers/core/reid/data/base.py
```diff
@@ -0,0 +1,78 @@
...
+        transforms (Optional[Compose]): Optional image transformations to apply
+        tracker_ids (list[str]): List of all unique tracker IDs in the dataset
+    """
+
+    def __init__(
+        self,
+        tracker_id_to_images: dict[str, list[str]],
+        transforms: Optional[Compose] = None,
+    ):
+        self.tracker_id_to_images = tracker_id_to_images
```
SkalskiP: We should validate the following conditions here:

- Whether there are any `tracker_ids` with fewer than 2 unique images — in that case, we won’t be able to select a positive example.
- Whether `self.tracker_id_to_images` contains more than one unique `tracker_id` — otherwise, we won’t be able to select a negative example.

trackers/core/reid/data/market_1501.py
```diff
@@ -0,0 +1,79 @@
...
+    to create a dictionary mapping tracker IDs to lists of image paths.
+
+    Args:
+        data_dir (str): The path to the Market1501 dataset.
+        split (str): The mode to use. Must be one of "train" or "test".
+
+    Returns:
+        Dict[str, List[str]]: A dictionary mapping tracker IDs to lists of image paths.
+    """
+    data_dir = os.path.join(data_dir, f"bounding_box_{split}")
```
SkalskiP: I think we should just pass this path as argument to `parse_market1501_dataset`

trackers/core/reid/data/market_1501.py
```diff
@@ -0,0 +1,79 @@
...
+    num_train_samples = len(tracker_ids) * (1 - validation_split_fraction)
+    train_tracker_ids = tracker_ids[: int(num_train_samples)]
+    validation_tracker_ids = tracker_ids[int(num_train_samples) :]
+    train_tracker_id_to_images = {
+        tracker_id: tracker_id_to_images[tracker_id] for tracker_id in train_tracker_ids
+    }
+    validation_tracker_id_to_images = {
+        tracker_id: tracker_id_to_images[tracker_id]
+        for tracker_id in validation_tracker_ids
+    }
```
SkalskiP: I think we should add `.split` method to `TripletsDataset` and move this logic there. Take a look at [`sv.DetectionDataset`](https://github.com/roboflow/supervision/blob/8b92485b6a9dcdb6d59a3e5f380f90728bfbe4d0/supervision/dataset/core.py#L179) for reference.

trackers/core/reid/data/market_1501.py
```diff
@@ -0,0 +1,79 @@
...
+    unique_ids = set(
+        os.path.basename(image_file).split("_")[0] for image_file in image_files
+    )
+    tracker_id_to_images: Dict[str, List[str]] = {
+        tracker_id: [] for tracker_id in unique_ids
+    }
+    for image_file in image_files:
+        tracker_id = os.path.basename(image_file).split("_")[0]
+        tracker_id_to_images[tracker_id].append(image_file)
+    return tracker_id_to_images
```
SkalskiP:
```suggestion
    tracker_id_to_images = defaultdict(list)
    
    for image_file in image_files:
        tracker_id = os.path.basename(image_file).split("_")[0]
        tracker_id_to_images[tracker_id].append(image_file)
    
    return dict(tracker_id_to_images)
```

trackers/core/reid/__init__.py
```diff
@@ -0,0 +1,3 @@
+from trackers.core.reid.data.market_1501 import get_market1501_dataset
+
+__all__ = ["get_market1501_dataset"]
```
SkalskiP: `TripletsDataset` should also be part of public API

trackers/utils/data_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        source_zip_path (str): The path to the zip file.
+        target_dir_path (str): The directory to extract the contents to.
+            If the directory doesn't exist, it will be created.
+    """
+    try:
+        with zipfile.ZipFile(source_zip_path, "r") as zip_ref:
+            zip_ref.extractall(target_dir_path)
+        print(f"Successfully extracted '{source_zip_path}' to '{target_dir_path}'")
+    except FileNotFoundError:
+        print(f"Error: Zip file '{source_zip_path}' not found.")
```
PawelPeczek-Roboflow: [1] Please raise error to the upstream

trackers/utils/data_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    except zipfile.BadZipFile:
+        print(f"Error: '{source_zip_path}' is not a valid zip file or is corrupted.")
```
PawelPeczek-Roboflow: [1]

trackers/utils/data_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    except Exception as e:
+        print(f"An error occurred: {e}")
```
PawelPeczek-Roboflow: [1]

PawelPeczek-Roboflow (changes requested):
Please avoid silent error handling

SkalskiP (changes requested):

tests/test_reid_dataset.py
SkalskiP: The proposed test is quite slow—it downloads a 150MB file every time it runs and only checks a single specific case. I’d advocate for removing it entirely. But if we decide to keep it, we should:

- Move it to a pre-release or post-merge test suite, rather than running it on every commit.
- Refactor the setup into a fixture and ensure the download is cached during local runs.

Instead, I’d focus on improving coverage of the underlying logic using fast unit tests, for example:

- Does the `TripletsDataset` constructor validate inputs correctly?
- Do `TripletsDataset.__getitem__` and `_get_triplet_image_paths` behave as expected with dummy data?
- Does `parse_market1501_dataset` parse the dataset correctly?

Use pytest.mark.parametrize where applicable. You can find some reference examples [here](https://github.com/roboflow/supervision/tree/develop/test/dataset).

trackers/core/reid/data/market_1501.py
```diff
@@ -0,0 +1,58 @@
...
+from collections import defaultdict
+from glob import glob
+from typing import Dict, List, Optional, Tuple, Union
+
+from torchvision.transforms import Compose
+
+from trackers.core.reid.data.base import TripletsDataset
+
+
+def parse_market1501_dataset(data_dir: str, split: str) -> Dict[str, List[str]]:
```
SkalskiP: `split` argument is unused

trackers/core/reid/data/market_1501.py
```diff
@@ -0,0 +1,58 @@
...
+        random_state (Optional[Union[int, float, str, bytes, bytearray]]): The random
+            state to use for the split.
+        shuffle (bool): Whether to shuffle the dataset.
+        transforms (Optional[Compose]): The transforms to apply to the dataset.
+
+    Returns:
+        Tuple[TripletsDataset, TripletsDataset]: A tuple of training and validation
+            `TripletsDataset` objects.
+    """
+    tracker_id_to_images = parse_market1501_dataset(data_dir, "train")
```
SkalskiP: `train` value is redundant

trackers/core/reid/data/base.py
```diff
@@ -0,0 +1,96 @@
...
+        negative_image = self._load_and_transform_image(negative_image_path)
+
+        return anchor_image, positive_image, negative_image
+
+    def split(
+        self,
+        split_ratio: float = 0.8,
+        random_state: Optional[Union[int, float, str, bytes, bytearray]] = None,
+        shuffle: bool = True,
+    ) -> Tuple[Dataset, Dataset]:
```
SkalskiP: should the output type be `Tuple[TripletsDataset, TripletsDataset]` ?

trackers/core/reid/data/base.py
```diff
@@ -0,0 +1,96 @@
...
+        train_tracker_id_to_images, validation_tracker_id_to_images = train_test_split(
+            self.tracker_id_to_images,
+            split_ratio=split_ratio,
```
SkalskiP: wrong argument name; it should be `train_ratio`

trackers/core/reid/data/base.py
```diff
@@ -0,0 +1,96 @@
...
+        return anchor_image, positive_image, negative_image
+
+    def split(
+        self,
+        split_ratio: float = 0.8,
+        random_state: Optional[Union[int, float, str, bytes, bytearray]] = None,
+        shuffle: bool = True,
+    ) -> Tuple[Dataset, Dataset]:
+        train_tracker_id_to_images, validation_tracker_id_to_images = train_test_split(
+            self.tracker_id_to_images,
```
SkalskiP: wrong argument type; it myst be list

.github/workflows/uv-test.yml
```diff
@@ -32,5 +32,5 @@ jobs:
       - name: 🧪 Run the Import test
         run: uv run python -c "import trackers"
 
-      #- name: 🧪 Run the Test
-        #run: uv run pytest
+      # - name: 🧪 Run the Test
```
SkalskiP: ?

SkalskiP (changes requested):

trackers/core/reid/data/base.py
soumik12345: Making this `TripletsDataset` causes the following error:
```
F821 Undefined name `TripletsDataset`
```

SkalskiP: @soumik12345 Have you tried adding `from __future__ import annotations` at the top of the file?

soumik12345: This worked. Thanks!

tests/test_reid_dataset.py
SkalskiP: @soumik12345 this comment is still pending

PawelPeczek-Roboflow (approved):

PR #40, opened 2025-05-05
Feat/core/bytetrack

AlexBodner:
# Description

This pull request adds the ByteTrack tracker to the list of implemented trackers.

No extra depedencies are required for usage, for testing I used pytest which is in the dev dependencies. The test file can be removed if you consider so, let me know.

## Type of change

-   [ ] New feature (non-breaking change which adds functionality)

The implementation has been thoroughly tested, passing all the evaluations, and showing the expected results.

The new tracker can be used in the following Google Colab, which shows 2 different examples of the Tracker in its 2 possible modes: using appearance features for high confidence boxes and using IoU.

There is also a test file in the route: trackers/core/bytetrack/tests_bytetrack.py, this tests cover the main functioning of the tracker and check that it behaves as expected.

Check trackers/__init__.py so that it can be imported correctly.
I've updated my trackers/core/deepsort/feature_extractor.py to be exactly like the one in the main branch, but for a reason it compares it to the old one.


trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,379 @@
...
+                new_tracker = ByteTrackKalmanBoxTracker(
+                    bbox=detection_boxes[detection_idx], feature=feature
+                )
+                self.trackers.append(new_tracker)
+
+            new_det = deepcopy(detections[detection_idx : detection_idx + 1])
+            new_det.tracker_id = np.array([-1])
+            updated_detections.append(new_det)
+
+        self.trackers = get_alive_trackers(
```
rolson24: Maybe this should live in the `update` function because `get_alive_trackers` is where we kill old lost tracks

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,379 @@
...
+                    feature = detection_features[detection_idx]
+
+                new_tracker = ByteTrackKalmanBoxTracker(
+                    bbox=detection_boxes[detection_idx], feature=feature
+                )
+                self.trackers.append(new_tracker)
+
+            new_det = deepcopy(detections[detection_idx : detection_idx + 1])
+            new_det.tracker_id = np.array([-1])
+            updated_detections.append(new_det)
```
rolson24: I know this works because `updated_detections` gets passed by reference to this function. But maybe its more explicit to return it?

rolson24 (commented):
The implementation looks pretty good! Just have a few comments about organization, but they are just my opinions.

trackers/core/bytetrack/kalman_box_tracker.py
```diff
@@ -0,0 +1,173 @@
+from typing import Optional, Union
+
+import numpy as np
+
+
+class ByteTrackKalmanBoxTracker:
```
soumik12345, line 6: This class seems an exact copy of `DeepSORTKalmanBoxTracker`. I think it would be better if we decoupled the `SORTKalmanBoxTracker` and`DeepSORTKalmanBoxTracker` into a separate module called `kalman_box_tracker` to make sure that we don't have tightly coupled class dependencies between the individual trackers while not repeating redundant code.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,379 @@
...
+            must be tracked before it is considered a 'valid' track. Increasing
+            `minimum_consecutive_frames` prevents the creation of accidental tracks
+            from false detection or double detection, but risks missing shorter
+            tracks. Before the tracker is considered valid, it will be assigned
+            `-1` as its `tracker_id`.
+        minimum_iou_threshold (float): IOU threshold for associating detections to
+            existing tracks.
+        high_prob_boxes_threshold (float): threshold for assigning predicted boxes to high probability class.
+            ByteTrack Specific parameter.
+        feature_extractor (FeatureExtractor OR THE NAME OF THE FINAL CLASS): model that will be use for extracting RE-ID features for
```
soumik12345: > FeatureExtractor OR THE NAME OF THE FINAL CLASS

What does this mean?

trackers/core/bytetrack/example_usage.py
```diff
@@ -0,0 +1,73 @@
...
+    "<path-to-your-video>.mp4"  # eg: SOURCE_VIDEO_PATH = "traffic_video_1.mp4"
+)
+
+TARGET_VIDEO_PATH = "result_video.mp4"
+sv.process_video(
+    source_path=SOURCE_VIDEO_PATH,
+    target_path=TARGET_VIDEO_PATH,
+    callback=callback,
+    show_progress=True,
+)
```
SkalskiP: @AlexBodner Let's not add example scripts like this directly to the repository. Instead, please:

- Create a Google Colab notebook based on [How to Track Objects with SORT Tracker](https://github.com/roboflow-ai/notebooks/blob/main/notebooks/how-to-track-objects-with-sort-tracker.ipynb) and [How to Track Objects with DeepSORT Tracker](https://github.com/roboflow-ai/notebooks/blob/main/notebooks/how-to-track-objects-with-deepsort-tracker.ipynb), and include your example there. For now install `trackers` from your branch (`pip install git+https://github.com/roboflow/trackers.git@feat/core/bytetrack` should work). We will marge it into [notebooks](https://github.com/roboflow/notebooks) repo before release.

- Add a new file at trackers/core/bytetrack/tracker.md, similar to the existing
[trackers/core/sort/tracker.md](https://github.com/roboflow/trackers/blob/main/docs/trackers/core/sort/tracker.md) and [trackers/core/deepsort/tracker.md](https://github.com/roboflow/trackers/blob/main/docs/trackers/core/deepsort/tracker.md). Fill it out in a similar way.

- Update the [mkdocs.yml](https://github.com/roboflow/trackers/blob/main/mkdocs.yml) file to link ByteTRACK, following the same pattern used for SORT and DeepSORT.

- Finally, test locally to ensure the documentation builds correctly.

trackers/core/bytetrack/tracker.py
AlexBodner: Should've ended being DeepSORT feature extractor for now. But that was because one of our talks where we said that the Feature Extractor will be independent of DeepSORT. So that was a comment for changing it if we changed that

trackers/core/bytetrack/tracker.py
AlexBodner: got it, will change that for more clarity

trackers/core/bytetrack/kalman_box_tracker.py
AlexBodner: It has some slight modifications, for example in the update function, but it is the same in essence. I totally agree in refactoring that to an additional module,  but for the moment I did it like this to be completely independent.

trackers/__init__.py
```diff
@@ -6,10 +6,11 @@
 logger = get_logger(__name__)
 
 try:
+    from trackers.core.bytetrack.tracker import ByteTrackTracker
```
SkalskiP: Please place from `trackers.core.bytetrack.tracker import ByteTrackTracker` inside its own `try`/`except` block with a dedicated warning message.

trackers/core/bytetrack/tracker.py
AlexBodner: Got it, will change that for more clarity

trackers/core/bytetrack/tracker.py
soumik12345: For now, let's keep this as `DeepSORTFeatureExtractor`.
We can change this in future when the ReID module that I'm working on is shipped.

trackers/core/bytetrack/example_usage.py
```diff
@@ -0,0 +1,73 @@
+import supervision as sv
```
soumik12345: I think that its better to remove this file from the PR. A sample code snippet would live in the corresponding docs page, but there's no need for a separate usage script.

trackers/core/bytetrack/tests_bytetrack.py
```diff
@@ -0,0 +1,375 @@
+import numpy as np
```
SkalskiP: @AlexBodner drop unit tests for now. We need to come up with good trackers testing strategy.

SkalskiP (changes requested):

trackers/core/bytetrack/tracker.py
AlexBodner: wdyt @soumik12345 ?

trackers/core/bytetrack/example_usage.py
AlexBodner: ok, removed

trackers/core/bytetrack/tracker.py
soumik12345: I agree with @rolson24; the `updated_detections` should be returned. Also, I'm a little unsure regarding the purpose of `_spawn_new_trackers`, shouldn't it update `self.trackers` with active/alive trackers?

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,376 @@
...
+        matched_indices, unmatched_trackers, unmatched_high_prob_detections = (
+            self._similarity_step(
+                high_prob_detections,
+                self.trackers,
+                self.high_prob_association_metric,
+                detection_features,
+            )
+        )
+
+        # Update matched trackers with assigned detections. !!Maybe it can be done casting matched_indices to np.array and using map!!
```
soumik12345: Do you mean something like the following?

```python
def _update_detections(
        self,
        trackers: list[ByteTrackKalmanBoxTracker],
        detections: sv.Detections,
        updated_detections: list[sv.Detections],
        matched_indices: list[tuple[int, int]],
    ):
        # Update matched trackers with assigned detections.
        det_bboxes = detections.xyxy

        def process_single_match(match_pair: tuple[int, int]) -> sv.Detections:
            row, col = match_pair
            t = trackers[row]
            t.update(det_bboxes[col])

            if t.tracker_id == -1:
                t.tracker_id = ByteTrackKalmanBoxTracker.get_next_tracker_id()
            
            new_det_slice = deepcopy(detections[col : col + 1])
            new_det_slice.tracker_id = np.array([t.tracker_id])
            return new_det_slice

        if matched_indices:
            newly_processed_detections = list(map(process_single_match, matched_indices))
            updated_detections.extend(newly_processed_detections)
```

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,376 @@
...
+        Clears all active tracks and resets the track ID counter.
+        """
+        self.trackers = []
+        ByteTrackKalmanBoxTracker.count_id = 0
+
+    def _get_appearance_distance_matrix(
+        self,
+        detection_features: np.ndarray,
+        trackers: list[ByteTrackKalmanBoxTracker],
+        #!!!In the future  this should be outside of the class in a utils file, as is reutilized in DeepSORT and here
```
soumik12345: This is an excellent observation. We need to figure out a better way to refactor the utilities that are used across multiple trackers.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,376 @@
...
+            unmatched_high_prob_detections,
+            updated_detections,
+        )
+
+        # Kill lost tracks
+        self.trackers = get_alive_trackers(
+            trackers=self.trackers,
+            maximum_frames_without_update=self.maximum_frames_without_update,
+            minimum_consecutive_frames=self.minimum_consecutive_frames,
+        )
```
soumik12345, line 222: I think it's better if the pruning logic is moved inside `_spawn_new_trackers`. This is how we handled it for the `SORTTracker`.

@rolson24, which way do you think is more logically coherent?

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,376 @@
...
+            for track activation. Only detections with confidence above this
+            threshold will create new tracks. Increasing this threshold
+            reduces false positives but may miss real objects with low confidence.
+        minimum_consecutive_frames (int): Number of consecutive frames that an object
+            must be tracked before it is considered a 'valid' track. Increasing
+            `minimum_consecutive_frames` prevents the creation of accidental tracks
+            from false detection or double detection, but risks missing shorter
+            tracks. Before the tracker is considered valid, it will be assigned
+            `-1` as its `tracker_id`.
+        minimum_iou_threshold (float): IOU threshold for associating detections to
```
soumik12345: Please mention what would be the effect of increasing and decreasing this value.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,376 @@
...
+            existing tracks.
+        high_prob_boxes_threshold (float): threshold for assigning predicted boxes to high probability class.
```
soumik12345: Please mention what would be the effect of increasing and decreasing this value.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,376 @@
...
+        self,
+        lost_track_buffer: int = 30,
+        frame_rate: float = 30.0,
+        track_activation_threshold: float = 0.25,
+        minimum_consecutive_frames: int = 3,
+        minimum_iou_threshold: float = 0.2,
+        high_prob_boxes_threshold: float = 0.5,
+        feature_extractor: Optional[DeepSORTFeatureExtractor] = None,
+        distance_metric: str = "cosine",  # None, 'cosine' or 'euclidean'
+        max_appearance_distance: float = 0.75,  # Default maximum distance for RE-ID matching with cosine distance.
```
soumik12345: Remove these comments.

soumik12345 (changes requested):
Hi @AlexBodner, thanks a lot for this PR!
The implementation of `ByteTrackTracker` seems correct as per the [paper](https://arxiv.org/pdf/2110.06864).

I have left some feedback, please address them and also fix the pre-commit failures.

trackers/core/bytetrack/tracker.py
AlexBodner: Raif told that he thought that It was more explícit to do this in update instead of doing it in spawn new trackers

rolson24: I honestly am not sure. But I think that you are right Soumik that if we put the pruning logic in `_spawn_new_trackers` in `SORTTracker`, then it probably makes sense to do the same for `ByteTracker`

trackers/core/bytetrack/tracker.py
AlexBodner: It could be. In my experience using numpy map tends to result faster than iterating lists.

trackers/core/bytetrack/tracker.py
AlexBodner: IMO is not that clear if we call a function spawn new trackers and inside of it do both the spawning and the killing. 
If we want to mantain a consistency between trackers it would be nice to keep it the same, but what seems more correct is to rather rename the function of move it directly to update like here.

trackers/core/bytetrack/example_usage.py
AlexBodner: Done, we would need to create a custom roboflow link to the google colab.

trackers/core/bytetrack/tracker.py
soumik12345: I think we all agree that decoupling the pruning logic from `_spawn_new_trackers` makes it logically coherent. Let's keep it like that in Bytetrack, I would make a PR to align SORT and DeepSORT.

trackers/core/bytetrack/tracker.py
soumik12345: Feel free to change to map.

trackers/core/bytetrack/tracker.py
soumik12345: https://github.com/roboflow/trackers/pull/48

soumik12345 (commented):
LGTM ✅

docs/trackers/core/bytetrack/tracker.md
```diff
@@ -0,0 +1,157 @@
...
+    model = RFDETRBase(device="cuda")  # Load the Object Detector
+    feature_extractor = DeepSORTFeatureExtractor.from_timm(
+        model_name="mobilenetv4_conv_small.e1200_r224_in1k",
+    )
+    tracker = ByteTrackTracker(feature_extractor=feature_extractor)
+
+    annotator = sv.LabelAnnotator(text_position=sv.Position.CENTER)
+
+    def callback(frame, _):
+        # Obtain bounding box predictions from RF-DETR
```
SkalskiP: please remove comments from code snippets; I'd like to keep them as short as possible

docs/trackers/core/bytetrack/tracker.md
```diff
@@ -0,0 +1,157 @@
...
+While calculating the appearance features with a Convolutional Neural Network might be slower than only comparing IoU, it makes it possible to track unpredictable trajectories and objects that dissapear and reappear in the scene, while also being suitable for real time tracking.
+
+BytTrack is independent on the object detector and feature extractor network so it can be used for any tracking task as long as the user provides the adequate model.
+
+
+## Examples
+
+=== "rf-detr"
+
+    ```python hl_lines="2 5-8 15"
```
SkalskiP: `2 5-8 15` specifies which lines of the code snippet should be highlighted. You can’t just copy and paste it from a different example, as it likely won’t match your own snippet. That’s why it’s important to build the docs locally and verify.

Same goes for rest of your snippets. 

Screenshot 2025-05-09 at 10 42 59

.github/workflows/mypy-test.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+          # TODO(https://github.com/astral-sh/setup-uv/issues/226): Remove this.
+          prune-cache: ${{ matrix.os != 'windows-latest' }}
+
+      - name: 🚀 Install Packages
+        run: uv pip install -r pyproject.toml --group dev --group docs --extra cpu --extra deepsort
+
+      - name: 🧪 Run MyPy Test
+        run: uv run mypy --config-file mypy.ini --exclude "tests|docs" .
```
SkalskiP: First batch of unit tests will come from here https://github.com/roboflow/trackers/tree/feat/reid-engine. Looks like we should exclude `test` directory not `tests`.

docs/trackers/core/bytetrack/tracker.md
```diff
@@ -0,0 +1,157 @@
...
+    import supervision as sv
+    from rfdetr import RFDETRBase
+    from trackers import ByteTrackTracker
+    from trackers.core.deepsort.feature_extractor import DeepSORTFeatureExtractor
+
+    model = RFDETRBase(device="cuda")  # Load the Object Detector
+    feature_extractor = DeepSORTFeatureExtractor.from_timm(
```
SkalskiP: It looks like we're using `DeepSORTFeatureExtractor` with both DeepSORT and ByteTrack. I think we should move it out of the `deepsort` directory and place it in a shared/common module. And of course, we should also rename it accordingly.

@soumik12345 Do you see anything that might prevent us from making this change?

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def __init__(
+        self,
+        lost_track_buffer: int = 30,
+        frame_rate: float = 30.0,
+        track_activation_threshold: float = 0.25,
+        minimum_consecutive_frames: int = 3,
+        minimum_iou_threshold: float = 0.2,
+        high_prob_boxes_threshold: float = 0.5,
+        feature_extractor: Optional[DeepSORTFeatureExtractor] = None,
```
SkalskiP: I'd like to keep the argument order consistent with `DeepSORTTracker`. This will make the `trackers` module easier to use, especially since many users don’t pass arguments by name. Let’s make it more user-friendly by keeping the argument order aligned wherever possible.

SkalskiP: Also, I noticed that the `feature_extractor` type differs here compared to `DeepSORTTracker`. I understand it's optional in this case, but aside from that, the type should be consistent.

@soumik12345 should we simplify the type in `DeepSORTTracker`, or update it here to match?

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,410 @@
...
+        distance_metric: str = "cosine",
+        max_appearance_distance: float = 0.75,
```
SkalskiP: @AlexBodner / @soumik12345 Is this the same as `appearance_threshold` in `DeepSORTTracker`?

SkalskiP (changes requested):

docs/trackers/core/bytetrack/tracker.md
onuralpszr: @SkalskiP because of that I include a commit that we need to install timm package plus I created a new sub pacakge for bytetrack but If person not using RE-ID it wont need timm package we also need to consider how we can make sure it is optional but not mandatory to use RE-ID feature in case If the person want to use IoU method and making sure minimal bytetrack can come with base package as well.

.github/workflows/mypy-test.yml
onuralpszr: @SkalskiP let me fix it.

onuralpszr: @SkalskiP  done in here.

docs/trackers/core/bytetrack/tracker.md
soumik12345: https://github.com/roboflow/trackers/pull/49 is going to move the feature extractor out of `deepsort` into the `reid` module.

trackers/core/bytetrack/tracker.py
soumik12345: Its going to be a simpler type called `ReIDModel` once we merge #49

trackers/core/bytetrack/tracker.py
soumik12345: Yes, they're basically the same.

SkalskiP: In that case @AlexBodner please make sure to keep arguments order and names consistant with other trackers especially `DeepSort`.

docs/trackers/core/bytetrack/tracker.md
AlexBodner: ok, didnt knew it did that

trackers/core/bytetrack/tracker.py
AlexBodner: @soumik12345 Should I leave it like this and once #49 is merged we change it?

trackers/core/bytetrack/tracker.py
AlexBodner: Changed!

docs/trackers/core/bytetrack/tracker.md
AlexBodner: Fixed!

trackers/core/bytetrack/tracker.py
AlexBodner: Now it is returned

SkalskiP:
Hi @onuralpszr — any chance you could take a look at these MyPy issues? They actually originate from @soumik12345’s PR: https://github.com/roboflow/trackers/pull/49. Your MyPy checks were merged in parallel, so this is the first PR where both changes are part of the same codebase.

I took a look at the errors but couldn’t figure out how to resolve them.

onuralpszr:
> Hi @onuralpszr — any chance you could take a look at these MyPy issues? They actually originate from @soumik12345’s PR: #49. Your MyPy checks were merged in parallel, so this is the first PR where both changes are part of the same codebase.
> 
> I took a look at the errors but couldn’t figure out how to resolve them.

Sure let me take a look.

trackers/core/reid/dataset/base.py
```diff
@@ -28,7 +34,7 @@ class TripletsDataset(Dataset):
     Attributes:
         tracker_id_to_images (dict[str, list[str]]): Dictionary mapping tracker IDs
             to lists of image paths
-        transforms (Optional[Compose]): Optional image transformations to apply
+        transforms (Union[Compose, ToTensor]): Image transformations to apply
```
SkalskiP, line 37: @onuralpszr what was the reason for making this change? The docstring no longer matches the actual type of the argument it describes.

trackers/core/reid/model.py
```diff
@@ -206,10 +206,23 @@ def _add_projection_layer(
...
+                    "cannot add projection layer."
+                )
+
+            num_model_features = getattr(self.backbone_model, "num_features")
+            if not isinstance(num_model_features, int):
+                raise TypeError(
+                    f"Backbone model's 'num_features' must be an int, "
+                    f"but got {type(num_model_features)}."
+                )
+
```
SkalskiP, line 223: @onuralpszr could you explain why this was needed and how it affected the MyPy check results?

onuralpszr: This part I wanted to be ensure we have that parameter (not related to mypy but cover the missing attribute error)

```python
            if not hasattr(self.backbone_model, "num_features"):
                raise AttributeError(
                    "Backbone model is missing 'num_features' attribute, "
                    "cannot add projection layer."
                )
```

Mypy part is here (we are not enforcing any so it is fine)

```python           
num_model_features = getattr(self.backbone_model, "num_features")
```

Here since I am getting any it will work but I wanted to make sure for "int" value so I added a check

```python
if not isinstance(num_model_features, int):
    raise TypeError(
        f"Backbone model's 'num_features' must be an int, "
        f"but got {type(num_model_features)}."
    )
    
```

trackers/core/reid/dataset/base.py
onuralpszr: It does match and Previous code already enforcing  "Compose, ToTensor"  in transforms  I just added type annotation 

https://github.com/roboflow/trackers/blob/b28d819ffe7000af79e1ead39478f09707940940/trackers/core/reid/dataset/base.py#L47

And for the function side you return "Tensor" and I made sure it is Tensor (which is I transformed) but IF* it is actually "Optional" I need to change small bit

trackers/core/reid/dataset/base.py
```diff
@@ -38,7 +44,7 @@ def __init__(
         transforms: Optional[Compose] = None,
     ):
         self.tracker_id_to_images = validate_tracker_id_to_images(tracker_id_to_images)
-        self.transforms = transforms or ToTensor()
+        self.transforms: Union[Compose, ToTensor] = transforms or ToTensor()
```
SkalskiP, line 47:
```suggestion
        self.transforms: Compose = transforms or transforms.Compose([ToTensor()])
```

@soumik12345 would it make sense to update this line like this? It might help unify the MyPy typing in this part of the code.

soumik12345: I think a better type hint would be `Union[Compose, Callable]` because any stateful callables should technically be possible to be composed as transforms.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,418 @@
...
+        detections.tracker_id = -np.ones(len(detections))
+        # Split into high confidence boxes and lower based on self.high_prob_boxes_threshold # noqa: E501
+        high_prob_detections, low_prob_detections = (
+            self._get_high_and_low_probability_detections(detections)
+        )
+
+        # If detector avaible, compute the features for high probability images
+        detection_features: Optional[np.ndarray]
+        if self.reid_model is not None:
+            print(len(high_prob_detections.xyxy))
```
soumik12345: @AlexBodner 
What's the need for this print statement?

AlexBodner: Oh i didnt push the deletion. Fixed

davidpagnon:
Very interested in this one :)

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+        high_prob_boxes_threshold (float): threshold for assigning predicted boxes to high probability class.
+            A higher value will classify only higher probability boxes as 'high probability'
+            per the ByteTrack algorithm, which are used in the first similarity step of
+            the algorithm.  If feature extractor is used, high probability boxes are the
+            only ones that are matched using the appearance features.
+    """  # noqa: E501
+
+    def __init__(
+        self,
+        reid_model: ReIDModel,
```
tstanczyk95: Originally, ByteTrack is designed to be used without re-ID. Re-ID is optional, as an extra feature for ByteTrack. As such, it would be better to make the parameter reid_model in the constructor default as None. It is better and more intuitive for the use to call:
tracker = ByteTrackTracker()
rather than
tracker = ByteTrackTracker(reid_model=None)

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+                t.tracker_id = ByteTrackKalmanBoxTracker.get_next_tracker_id()
+
+            new_det = deepcopy(detections[col : col + 1])
+            # Add cast to clarify type for mypy
+            new_det = cast(sv.Detections, new_det)  # ADDED cast
+            new_det.tracker_id = np.array([t.tracker_id])
+            updated_detections.append(new_det)
+        return updated_detections
+
+    def update(self, detections: sv.Detections, frame: np.ndarray) -> sv.Detections:
```
tstanczyk95: The frame param: same remark as for the reid_model param in the constructor. There should be set as default to None since the ByteTrack is originally intended to run without re-ID. If detections are provided, ByteTrack should be able to run without frames.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+            sv.Detections
+        ] = []  # List for returning the updated detections with its new assigned track id # noqa: E501
+
+        # Predict new locations for existing trackers
+        for tracker in self.trackers:
+            tracker.predict()
+        # Assign a default tracker_id with the correct shape
+        detections.tracker_id = -np.ones(len(detections))
+        # Split into high confidence boxes and lower based on self.high_prob_boxes_threshold # noqa: E501
+        high_prob_detections, low_prob_detections = (
```
tstanczyk95, line 155: Original ByteTrack detection split:

high conf dets: conf > args.track_thresh (0.6)
low conf dets: args.track_thresh > conf > 0.1

[Hence if conf == args.track_thresh: dets skipped... ---> I would say if equal then treat it as high conf dets]

About the track_thresh, yes demo_track.py includes 0.5 as the default, but for the actual evaluations they use 0.6: https://github.com/FoundationVision/ByteTrack/blob/main/tools/track.py#L105

Here however, in this pull request, we have this instead:
higher dets are above self.high_prob_boxes_threshold (currently set as 0.5), lower dets are ALL the other dets (even with conf <= 0.1)

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+        matched_indices = []
+        unmatched_trackers = set(range(len(trackers)))
+        unmatched_detections = set(range(len(detection_boxes)))
+
+        if len(trackers) > 0 and len(detection_boxes) > 0:
+            row_indices, col_indices = linear_sum_assignment(
+                similarity_matrix, maximize=True
+            )
+            for row, col in zip(row_indices, col_indices):
+                if similarity_matrix[row, col] >= min_similarity_thresh:
```
tstanczyk95, line 288: This is a bit concerning here:
Only discarding the matches based on minimum_iou_threshold rather than blocking similarity matrix entries from an association. 

Original ByteTrack passes the cost_threshold (equivalent of minimum_iou_threshold) directly to the linear solver:
cost, x, y = lap.lapjv(cost_matrix, extend_cost=True, cost_limit=thresh)

The pull request version on the other hand, firstly runs the linear assignment on all entries of the similarity matrix, no matter what, and only after obtaining the matches (matched indices), discards them based on the corresponding values in similarity matrix.

Original ByteTrack: blocking some entries (setting a high value there) before optimization. Then we might (not guaranteed) still get matches for all pairs.
This pull request version: matching done, then discarding matches if the threshold doesn't allow it - some matches are completely discarded...

The results might be different in the two cases above.

My intuition: it's better to block some match entries (ith tracklet with jth detection) before the optimization/matching process, e.g. by setting these entries to some high/low values above/below the threshold - rather than discard them after the matching is done.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+                - unmatched_detections_indices: A set of indices for detections
+                  that were not matched.
+
+        Raises:
+            Exception: If an unsupported `association_metric` is provided.
+        """  # noqa: E501
+        similarity_matrix = None
+        if association_metric == "IoU":
+            # Build IOU cost matrix between detections and predicted bounding boxes
+            similarity_matrix = get_iou_matrix(trackers, detections.xyxy)
```
tstanczyk95: Preliminary note: Original ByteTrack constructs a cost matrix (1-IoU) rather than similarity matrix (IoU). While it should not matter or cause any significant differences, it is good to keep it in mind when making the comaparizons to the original ByteTrack implementation. 

Mind that original ByteTrack does also this during its first and third association step:
if not self.args.mot20: # thus all the datasets except mot20
            dists = matching.fuse_score(dists, detections)
|
In short, it does this:
fuse_sim = iou_sim * det_scores
fuse_cost = 1 - fuse_sim
return fuse_cost

So basically it enforces matches for each detection based on detection confidence.

It might lead to slightly different results, especially in case of noisy detector. Stronger detections will basically dominate.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+        distance_metric (str): Distance metric for appearance features (e.g., 'cosine',
+            'euclidean'). See `scipy.spatial.distance.cdist`.
+        high_prob_boxes_threshold (float): threshold for assigning predicted boxes to high probability class.
+            A higher value will classify only higher probability boxes as 'high probability'
+            per the ByteTrack algorithm, which are used in the first similarity step of
+            the algorithm.  If feature extractor is used, high probability boxes are the
+            only ones that are matched using the appearance features.
+    """  # noqa: E501
+
+    def __init__(
```
tstanczyk95, line 62: Params as in the original ByteTrack should be:
* minimum_consecutive_frames = 2 (1 for the very first frame), NOT 3 ---> the evaluation script will penalize missing records over more frames.
* minimum_iou_threshold = 0.1, NOT 0.2 ---> ByteTrack allows lower IoU similarity (higher IoU cost) for the first assocation step: https://github.com/FoundationVision/ByteTrack/blob/main/tools/track.py#L107 (0.9). Setting more demanding value will limit the number of associations made compared to the original version.
* **EXTRA NOTE:** Mind that ByteTrack has 3 (thus not 2) association steps (one extra for newly born tracklets), and each of the 3 steps has a different matching threshold ---> see a separate comment on that.
* high_prob_boxes_threshold = 0.6, NOT 0.5: https://github.com/FoundationVision/ByteTrack/blob/main/tools/track.py#L105
* track_activation_threshold = high_prob_boxes_threshold + 0.1 = 0.7, NOT 0.25 <--- **This might cause chaos with non-oracle detections (coming from a detector, with confidences < 1.0)** - Also noisy detections will be considered for new tracklet creation, creating a noise and false positive tracklets.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+            high_prob_detections,
+            updated_detections,
+            matched_indices,
+            detection_features,
+        )
+
+        remaining_trackers = [self.trackers[i] for i in unmatched_trackers]
+
+        # Step 2: associate Low Probability detections with remaining trackers
+        matched_indices, unmatched_trackers, unmatched_detections = (
```
tstanczyk95: Original ByteTrack:

1st assoc: currently tracked + lost tracklets
2nd assoc: only currently tracked tracklets

Pull request version:
Both associations use both currently tracked tracklets AND lost tracklets
Basically, there is no clear division between these two. Only this state of a ByteTrackKalmanBoxTracker instance: self.time_since_update, yet not used in bytetrack/tracker.py

Including lost tracklets in the second association step, with the low confidence detections might cause chaos with associations and many redundant associations, which in turn will lower the accuracy performance. Especially in case of noisy detector/detections.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+                  that were not matched.
+
+        Raises:
+            Exception: If an unsupported `association_metric` is provided.
+        """  # noqa: E501
+        similarity_matrix = None
+        if association_metric == "IoU":
+            # Build IOU cost matrix between detections and predicted bounding boxes
+            similarity_matrix = get_iou_matrix(trackers, detections.xyxy)
+            thresh = self.minimum_iou_threshold
```
tstanczyk95, line 380: Original ByteTrack:
Each assoc step: different threshold (cost_limit, allowing IoU distance):
1st assoc.: args.match_thresh (0.9) (similarity: 0.1)
2nd assoc.: 0.5 (similarity: 0.5)
3rd assoc.: 0.7 (similarity: 0.3)

Pull request version:
always the same value: similarity thresh: 0.2
|
It will cause different results. The scale of difference might vary depending in the dataset and importantly, detection quality.

Basically, this pull request version allows more associations to happen than the original ByteTrack. Especially within the second association step with low confidence, thus more noisy, detections. Depending on the detector/detections, it might degrade the performance.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+            )
+            for row, col in zip(row_indices, col_indices):
+                if similarity_matrix[row, col] >= min_similarity_thresh:
+                    matched_indices.append((row, col))
+                    unmatched_trackers.remove(row)
+                    unmatched_detections.remove(col)
+
+        return matched_indices, unmatched_trackers, unmatched_detections
+
+    def _spawn_new_trackers(
```
tstanczyk95, line 295: **[ Big difference here ]**

Original ByteTrack:
unmatched high conf dets are initiated as new tracklets only if they are matched with the next frame's unmatched high conf detections - with dist threshold 0.7 (= similarity threshold 0.3). Unless it is the very first frame. Then the new tracklets are instantly activated (see the "MIND THAT:" remark below for more).
|
Thus also: 2 frames are enough, not 3. And based on the high confidence unmatched detections between two frames.

This pull request version:
(_spawn_new_trackers)
Create a new tracklet from unmatached high conf detection. Then include it in the same pool as currently tracked tracklets (active tracklets) for matches with high and low conf dets. (original: matching only with the unmatched high conf dets, separate 3rd assoc step).
|
This will cause different results. The scale of difference might vary depending on the dataset and importantly, detection quality. More noisy detections = more false positive tracklets created. Also more chaos when mixing these freshly created, yet not mature tracklets in the same pool with other tracklets as mentioned above.


**[ MIND THAT: ]**
In the original ByteTrack, the STrack's (tracklet representation) activate function (https://github.com/FoundationVision/ByteTrack/blob/main/yolox/tracker/byte_tracker.py#L45) does not activate the tracklet yet. It sets self.state = TrackState.Tracked.
Only the STrack's update function truly activates it by setting: self.is_activated = True (STrack's reactivate function also does it).
As a result, the newly born tracklets from frame t are activated only after being confirmed (matched to remaining high conf detections) at frame t+1
**---Exception:** new tracklets at the very first frame (in ByteTrack settings: t=1) - only then STrack's function truly activates the tracklet by setting self.is_activated = True

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+            trackers=self.trackers,
+            maximum_frames_without_update=self.maximum_frames_without_update,
+            minimum_consecutive_frames=self.minimum_consecutive_frames,
+        )
+        final_updated_detections: sv.Detections = sv.Detections.merge(
+            updated_detections
+        )
+        if len(final_updated_detections) == 0:
+            final_updated_detections.tracker_id = np.array([], dtype=int)
+        return final_updated_detections
```
tstanczyk95, line 228: [ Just a note ]
Originally provided with ByteTrack - just before returning the update's function final output - it removes some tracks per frame:

self.tracked_stracks, self.lost_stracks = remove_duplicate_stracks(self.tracked_stracks, self.lost_stracks)
(https://github.com/FoundationVision/ByteTrack/blob/main/yolox/tracker/byte_tracker.py#L285)

However, it, might cause little bugs with tracklet management. Therefore, I just mention it exists, but I don't recommend adding it.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+        # consistent time-based tracking across different frame rates.
+        self.maximum_frames_without_update = int(frame_rate / 30.0 * lost_track_buffer)
+        self.minimum_consecutive_frames = minimum_consecutive_frames
+        self.minimum_iou_threshold = minimum_iou_threshold
+        self.track_activation_threshold = track_activation_threshold
+        self.high_prob_boxes_threshold = high_prob_boxes_threshold
+        self.high_prob_association_metric = "IoU" if reid_model is None else "RE-ID"
+        self.reid_model = reid_model
+        self.distance_metric = distance_metric
+        self.trackers: list[ByteTrackKalmanBoxTracker] = []
```
tstanczyk95: I would rather call it "tracks", not "trackers". Tracker is an instance that tracks entities (referred to as tracks or opt. tracklets)

tstanczyk95 (commented):
Overall: nicely written. It looks clear and intuitive. Very good work here.

I spotted quite some differences compared to the original ByteTrack implementation from here: https://github.com/FoundationVision/ByteTrack/tree/main and pointed them with my remarks. Let me know in case you need extra information or clarifications.

Note: I focused on the non-re-ID version as this os the primary focus of ByteTrack. There exist other SORT-based algorithms explicitly using re-ID.

trackers/core/bytetrack/tracker.py
```diff
@@ -0,0 +1,432 @@
...
+from trackers.core.base import BaseTrackerWithFeatures
+from trackers.core.bytetrack.kalman_box_tracker import ByteTrackKalmanBoxTracker
+from trackers.core.reid import ReIDModel
+from trackers.utils.sort_utils import (
+    get_alive_trackers,
+    get_iou_matrix,
+)
+
+
+class ByteTrackTracker(BaseTrackerWithFeatures):
```
tstanczyk95, line 18: [ **Extra step after tracking** ]

Mind that in the original ByteTrack implementation, there is also a post-processing interpolation step:
https://github.com/FoundationVision/ByteTrack/blob/main/tools/interpolation.py .
You can see it to be called here: https://github.com/FoundationVision/ByteTrack?tab=readme-ov-file#tracking .
It can be run on a single or multiple tracking text outputs.

It basically smooths the bounding box coordinates over the trajectories of each tracklet. A detector might provide a bit of noise, i.e. "jumping" /"shaky" bounding boxes of an object over the consecutive frames. This interpolation step rectifies it a bit, so that the overall trajectories are smoother. 

In case of MOT17 and MOT20 it did improve the final results (HOTA, MOTA, IDF1; sometimes even by a few points).

Note two things:
* You do not need to evaluate mota during this process, this lines 125 and 127 are redundant:
https://github.com/FoundationVision/ByteTrack/blob/main/tools/interpolation.py#L125
Hence there is no need for any dataset at this step.
* For actual evaluations, it is highly **NOT** recommended to use the motmetrics package as done here: https://github.com/FoundationVision/ByteTrack/blob/main/tools/interpolation.py#L4, but use the official evaluation tool: TrackEval https://github.com/JonathonLuiten/TrackEval
TrackEval is an official evaluation script in the community, while motmetrics produce different results, to a higher or lower extend, resulting in confusion and ambiguities.

tstanczyk95 (commented):
Additional remark to my last ByteTrack review. It discusses additional functionality and post-processing step from the original ByteTrack implementation.

trackers/core/bytetrack/tracker.py
AlexBodner: Solved

trackers/core/bytetrack/tracker.py
AlexBodner: Good catch, in the paper they dont mention the lower bound threshold, but i could find it in the implementation:
https://github.com/FoundationVision/ByteTrack/blob/main/yolox/tracker/byte_tracker.py
lines 177-182

AlexBodner: Should we include this 0.1 as a parameter? I think so, but bytetrack hardcodes it as 0.1

trackers/core/bytetrack/tracker.py
AlexBodner: I see the difference, do you think that I should patch the implementation with scipy by adding the big costs or should we switch to lapjv?

trackers/core/bytetrack/tracker.py
AlexBodner: Changed defaults

trackers/core/bytetrack/tracker.py
AlexBodner: it probably would result in a faster implementation with lapjv, but it can create dependency issues, specially for installing in some devices which might not be compatible with the build for this package, so im inclined towards fixing the scipy one

trackers/core/bytetrack/tracker.py
AlexBodner: Im adapating the code to work in this style. Should we fuse scores when working with RE-ID features? I could only find this example where they did actually used embeddings for matching:
https://github.com/FoundationVision/ByteTrack/blob/main/tutorials/qdtrack/tracker_reid_motion.py
here they use the `fuse_motion` function, and also the script is called tracker_reid_motion which sounds like they are doing something different to ByteTrack, but they implement a class of ByteTrack there. Is there any other implementation you could point me to?

trackers/core/bytetrack/tracker.py
AlexBodner: changed it for `        remaining_trackers = [self.trackers[i] for i in unmatched_trackers if self.trackers[i].time_since_update == 1]`. This is enough, to fix the behaviour difference with lost tracks, right? Even though it might be nicer to explictly called them lost tracks and treat them differently

trackers/core/bytetrack/tracker.py
AlexBodner: got it, we can add it as another parameter to the tracker. Didnt see it in the paper as well, but is clear in the source code

trackers/core/bytetrack/tracker.py
AlexBodner: Fixed in commit
https://github.com/roboflow/trackers/commit/f1c440f1ae6da12e7b152786d15bd7120674383f

trackers/core/bytetrack/tracker.py
AlexBodner: Done

trackers/core/bytetrack/tracker.py
AlexBodner: Ok, wont change that for the moment

trackers/core/bytetrack/tracker.py
AlexBodner: Done
https://github.com/roboflow/trackers/commit/cd709d468020ed4ae88080bb3bda88a335269fb9

trackers/core/bytetrack/tracker.py
AlexBodner: Done in:
https://github.com/roboflow/trackers/commit/468f49a261776166fc02fce59b1ef227d960c2bc

trackers/core/bytetrack/tracker.py
tstanczyk95: We could include it as a parameter with the default value of 0.1. Usually, confidence of 0.1 or below means quite unreliable and noisy detections. However, we can leave the user an option to adjust it as, depending on their scene environment, they might have very low confidence detections in general (e.g. blurry images, night scenarios, etc.).

trackers/core/bytetrack/tracker.py
tstanczyk95: > it probably would result in a faster implementation with lapjv, but it can create dependency issues, specially for installing in some devices which might not be compatible with the build for this package, so im inclined towards fixing the scipy one

In this case, I would also opt for the SciPy version.

trackers/core/bytetrack/tracker.py
tstanczyk95: From my point of view, we should leave the subject of re-ID within the ByteTrack and come back to it with other (possibly related or derived) trackers which are actually designed to use it. ByteTrack original intention was to be used without re-ID, plain IoU. When someone wants a tracker with re-ID, they look for something else than ByteTrack.

trackers/core/bytetrack/tracker.py
tstanczyk95: Given the current implementation, it seems it is enough. However, we could run it on a few sequences, verify particular cases (e.g. with occlusion) and log the tracklet numbers to be sure it behaves as expected.

trackers/core/bytetrack/tracker.py
tstanczyk95: > Fixed in commit [f1c440f](https://github.com/roboflow/trackers/commit/f1c440f1ae6da12e7b152786d15bd7120674383f)

Looks good! As with one of the previous comments, to be sure, I recommend running it on a few sequences and log the tracklet numbers to be see if it behaves as expected.

trackers/core/bytetrack/tracker.py
AlexBodner: Implemented

trackers/core/bytetrack/tracker.py
tstanczyk95: > Done in: [468f49a](https://github.com/roboflow/trackers/commit/468f49a261776166fc02fce59b1ef227d960c2bc)

I am a bit concerned about the lines 52-58. It fills the extended cost matrix with values of the initial one, with cost_limit/2.0 and with 0s. I am not sure how it is supposed to help scipy.optimize.linear_sum_assignment to block the entries which have to be excluded by the threshold.

E.g. if we have 3 tracklets and 5 detections and there is cost_limit different than infinity, an extended cost matrix as in the image below will be created:

Selection_273

I don't see how it should involve the actual cost_limit to exclude the relevant entries from c11-c35.

Was it done like that or in a similar way in lap.lapjv? If so, do you have a link or reference to it?

AlexBodner: It is used here: https://github.com/gatagat/lap/blob/master/_lapjv_cpp/_lapjv.pyx

the things is that if a real cost (c_{ij}) is bigger than the cost limit, then, the algorithm will instead want to match the newly added rows and columns in between themselves and the total cost of that matches would be cost_limit/2 *2 = cost_limit, but if its more optimal to match to one of the costs smaller that cost_limit then it would do so and then match the dummies to one of cost 0

SkalskiP:
I'm closing this PR, as we already merged ByteTrack via https://github.com/roboflow/trackers/pull/174.

PR #49, opened 2025-05-09
feat(trackers): Implement `ReIDModel` for feature extraction and triplet training

soumik12345:
# Description

This PR implements the `ReIDModel`, a drop-in replacement for the `DeepSORTFeatureExtractor`. It is compatible with all trackers utilizing appearance features and provides additional features to train/fine-tune ReID models using the `TripletsDataset`.

## Example: Fine-tune a `timm` model

```python
from trackers import ReIDModel
from trackers.core.reid import get_market1501_dataset

from torch.utils.data import DataLoader


dataset = get_market1501_dataset(data_dir="datasets/reid/Market-1501-v15.09.15/bounding_box_train")
dataloader = DataLoader(dataset, batch_size=32, shuffle=True)

model = ReIDModel.from_timm(model_name="resnetv2_50.a1h_in1k")
model.train(
    dataloader,
    epochs=10,
    log_to_wandb=True,
    learning_rate=5e-4,
    projection_dimension=len(dataset),
    freeze_backbone=True,
    checkpoint_interval=2,
)
```
👉 [**Tensorboard run**](https://wandb.ai/19soumik-rakshit96/trackers/runs/lusw294y/tensorboard)
👉 [**Wandb run**](https://wandb.ai/19soumik-rakshit96/trackers/runs/0pyr20fs)

## Example: Load a fine-tuned checkpoint

```python
from trackers import ReIDModel

model = ReIDModel.from_timm("logs/checkpoints/reid_model_10.safetensors")
```

## Dependency changes

This PR introduces two new extras called `reid` and `metrics` that consists of the following dependencies:
```toml
reid = [
    "safetensors>=0.5.3",
    "timm>=1.0.15",
]

metrics = [
    "tensorboard>=2.19.0",
    "wandb>=0.19.11",
]
```

Additionally, the extra `deepsort` is removed and `"scipy>=1.13.1"` is moved to the main dependency list, assuming it's a common dependency for all trackers utilizing linear sum assignment.


pyproject.toml
```diff
@@ -36,7 +36,9 @@ keywords = ["tracking","mot","sort","deepsort", "machine-learning", "deep-learni
 dependencies = [
     "firerequests>=0.1.2",
     "numpy>=2.0.2",
+    "safetensors>=0.5.3",
     "supervision>=0.26.0rc7",
+    "tensorboard>=2.19.0",
     "validators>=0.34.0",
```
onuralpszr: Is this temporary and permanent ? because based on code the you also need to include torch too.

soumik12345: I think we should create a separate extra called `reid` and move these.

trackers/utils/data_utils.py
```diff
@@ -0,0 +1,56 @@
...
+        FileNotFoundError: If the zip file doesn't exist.
+        zipfile.BadZipFile: If the file is not a valid zip file or is corrupted.
+        Exception: If any other error occurs during extraction.
+    """
+    with zipfile.ZipFile(source_zip_path, "r") as zip_ref:
+        zip_ref.extractall(target_dir_path)
+    print(f"Successfully extracted '{source_zip_path}' to '{target_dir_path}'")
+
+
+def validate_tracker_id_to_images(
```
SkalskiP: This function only makes sense in the context of ReID, but that’s not mentioned anywhere — not in the function name, the docstring, the file path, or the package name.

We should either rename the function and update the docstring to clarify its ReID-specific purpose, or consider relocating it to a more appropriate module that reflects its use case.

trackers/core/reid/model.py
```diff
@@ -0,0 +1,360 @@
...
+
+from trackers.core.reid.callbacks import (
+    BaseCallback,
+    TensorboardCallback,
+    WandbCallback,
+)
+from trackers.utils.torch_utils import parse_device_spec
+
+
+class ReIDModel:
```
SkalskiP, line 85: given that we're adding `ReIDModel`, I think we should remove `DeepSORTFeatureExtractor` from the codebase.

pyproject.toml
```diff
@@ -76,9 +77,11 @@ rocm624 = [
...
 ]
 
-deepsort = [
-    "scipy>=1.13.1",
+reid = [
+    "safetensors>=0.5.3",
+    "tensorboard>=2.19.0",
     "timm>=1.0.15",
+    "wandb>=0.19.11",
 ]
```
SkalskiP: If we go this route, `tensorboard` and `wandb` will become de facto mandatory requirements for using ReID with trackers. What about users who just want to use a pre-trained or fine-tuned feature extraction model? 

RF-DETR solves this by using a separate dependency extra:

```toml
metrics = [
    "tensorboard>=2.13.0",
    "wandb"
]
```

I think we can do something similar.

docs/trackers/core/reid/model.md
```diff
@@ -0,0 +1,52 @@
...
+    model = ReIDModel.from_timm(**config["model_metadata"])
+    if config["projection_dimension"]:
+        model.add_projection_layer(
+            projection_dimension=config["projection_dimension"]
+        )
+
+    # Load state dict to the backbone model
+    for k, v in state_dict.items():
+        state_dict[k].to(model.device)
+    model.backbone_model.load_state_dict(state_dict)
```
SkalskiP: We need to simplify this API — users should be able to load a fine-tuned model the same way they load a pre-trained one.

```python
model = ReIDModel.from_timm("resnetv2_50.a1h_in1k")
```

```python
model = ReIDModel.from_timm("checkpoints/reid_model_10.safetensors")
```

I suggest renaming `model_name` to `model_checkpoint`, and moving the logic from this snippet inside the `from_timm` method. It should handle the validation internally and determine whether the input is a pre-trained or fine-tuned model.

mkdocs.yml
```diff
@@ -110,3 +110,6 @@ nav:
   - Trackers:
     - SORT: trackers/core/sort/tracker.md
     - DeepSORT: trackers/core/deepsort/tracker.md
+  - ReID:
+    - Datasets: trackers/core/reid/dataset.md
+    - Model: trackers/core/reid/model.md
```
SkalskiP: In general, I think we should merge the ReID models and ReID datasets into a single docs page. It would be much more helpful to show users the full workflow in one place: how to load a pre-trained model, use it for tracking, fine-tune it, and then load the fine-tuned version.

trackers/core/reid/model.py
```diff
@@ -0,0 +1,360 @@
...
+            "freeze_backbone": freeze_backbone,
+            "model_metadata": self.model_metadata,
+        }
+
+        # Initialize callbacks
+        callbacks: list[BaseCallback] = []
+        if log_to_tensorboard:
+            callbacks.append(TensorboardCallback())
+        if log_to_wandb:
+            callbacks.append(WandbCallback(config=config))
```
SkalskiP: We should make the imports for `TensorboardCallback` and `WandbCallback` optional, so that if `tensorboard` or `wandb` aren't installed, the user gets a clear warning suggesting they install the appropriate extra dependencies.

trackers/core/reid/callbacks.py
```diff
@@ -0,0 +1,77 @@
...
+        pass
+
+    def on_train_batch_end(self, logs: dict, idx: int):
+        pass
+
+    def on_validation_batch_start(self, logs: dict, idx: int):
+        pass
+
+    def on_validation_batch_end(self, logs: dict, idx: int):
+        pass
```
SkalskiP: Do we want to log on every batch or on every epoch?

trackers/core/reid/callbacks.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def on_checkpoint_save(self, checkpoint_path: str, epoch: int):
+        pass
+
+    def on_end(self):
+        pass
+
+
+class TensorboardCallback(BaseCallback):
```
SkalskiP, line 33: I think it would be cool to have one more callback that would save `matplotlib` chart as well.

SkalskiP (changes requested):

trackers/core/reid/callbacks.py
soumik12345: I like to log after every batch to get the raw insights.
If the logs per batch look too dense, we can add events to the callback called `on_epoch_end` where the accumulated metrics across the batch are logged.

soumik12345: Hi @SkalskiP, I updated the metric logger callbacks in the following manner:
- Only train loss is logged every batch step under the key `batch/train/loss`
- Train loss is also logged after every epoch under the key `train/loss`
- Validation loss is logged only after every epoch under the key `validation/loss`

Would love to know your thoughts on this.

trackers/core/reid/callbacks.py
```diff
@@ -0,0 +1,140 @@
+from typing import Any, Optional
+
+import matplotlib.pyplot as plt
+
+
+class BaseCallback:
```
SkalskiP, line 7: Is there any reason to believe we'll be training anything else within `trackers` besides ReID? I'm wondering whether these callbacks should live in the `reid` package or somewhere more general. If nothing else comes to mind for now, I’m fine with keeping them in `reid`.

trackers/core/reid/callbacks.py
```diff
@@ -0,0 +1,140 @@
...
+            path=checkpoint_path,
+            name=f"checkpoint_{self.run.id}",
+            aliases=[f"epoch-{epoch}", "latest"],
+        )
+
+    def on_end(self):
+        self.run.finish()
+
+
+class MatplotlibCallback(BaseCallback):
```
SkalskiP, line 107: Could you share an example of the result plot here?

trackers/core/deepsort/tracker.py
```diff
@@ -51,7 +47,7 @@ class DeepSORTTracker(BaseTrackerWithFeatures):
 
     def __init__(
         self,
-        feature_extractor: Union[DeepSORTFeatureExtractor, torch.nn.Module, str],
+        reid_model: ReIDModel,
```
SkalskiP, line 50: @AlexBodner we'd like to replicate the same API in your ByteTrack PR: https://github.com/roboflow/trackers/pull/40. I think you can rebase your branch onto `feat/reid-training` and start migrating to use `ReIDModel`.

trackers/core/deepsort/feature_extractor.py
```diff
@@ -1,169 +0,0 @@
...
-import supervision as sv
-import timm
-import torch
-import torch.nn as nn
-import torchvision.transforms as transforms
-
-from trackers.utils.torch_utils import parse_device_spec
-
-
-class FeatureExtractionBackbone(nn.Module):
```
SkalskiP, line 13: @soumik12345 we need to make sure that all content and examples using `DeepSORTFeatureExtractor` are migrated to `ReIDModel`.

pyproject.toml
```diff
@@ -76,11 +77,17 @@ rocm624 = [
     "pytorch-triton-rocm>=2.0.0",
 ]
 
-deepsort = [
-    "scipy>=1.13.1",
+reid = [
+    "safetensors>=0.5.3",
     "timm>=1.0.15",
 ]
```
SkalskiP, line 83: `deepsort/tracker.md` still contains instructions referencing the `deepsort` extra dependencies. It looks like we need to migrate those to `reid`.

trackers/core/reid/__init__.py
```diff
@@ -0,0 +1,5 @@
+from trackers.core.reid.dataset.base import TripletsDataset
+from trackers.core.reid.dataset.market_1501 import get_market1501_dataset
+from trackers.core.reid.model import ReIDModel
+
+__all__ = ["ReIDModel", "TripletsDataset", "get_market1501_dataset"]
```
SkalskiP: Shouldn't we use the same `try/except` import pattern here? There's no guarantee users will have the `reid` extras installed.

Also, I think all short imports should be centralized in one place — `trackers/__init__.py`. That’s the approach we took in `supervision`.

docs/trackers/core/deepsort/tracker.md
```diff
@@ -96,16 +90,14 @@ DeepSORT extends the original [SORT](../sort/tracker.md) algorithm by integratin
...
     import supervision as sv
-    from trackers import DeepSORTFeatureExtractor, DeepSORTTracker
+    from trackers import DeepSORTTracker, ReIDModel
     from transformers import RTDetrV2ForObjectDetection, RTDetrImageProcessor
 
-    feature_extractor = DeepSORTFeatureExtractor.from_timm(
-        model_name="mobilenetv4_conv_small.e1200_r224_in1k"
-    )
-    tracker = DeepSORTTracker(feature_extractor=feature_extractor)
+    reid_model = ReIDModel.from_timm("resnetv2_50.a1h_in1k")
```
SkalskiP, line 99: @soumik12345 I'm curious—what made you decide to switch the example `timm` model from `mobilenetv4_conv_small.e1200_r224_in1k` to `resnetv2_50.a1h_in1k`? Was there a specific reason for the change?

docs/trackers/core/reid/reid.md
```diff
@@ -0,0 +1,43 @@
+# ReID Model
```
SkalskiP: That header is no longer valid. We talk about a lot more than just `ReID Model`. We need to make it more general.

docs/trackers/core/reid/reid.md
```diff
@@ -0,0 +1,43 @@
...
+    from trackers import ReIDModel
+
+    model = ReIDModel.from_timm("logs/checkpoints/reid_model_10.safetensors")
+    ```
+
+::: trackers.core.reid.dataset.base.TripletsDataset
+
+::: trackers.core.reid.dataset.market_1501.get_market1501_dataset
+
+::: trackers.core.reid.model.ReIDModel
```
SkalskiP: I think we need to provide much more context here. We should start by introducing `ReIDModel`—explain what it is and how to load it. Then, clarify how it integrates with `timm`, and link to the full list of supported `timm` models. It would also be helpful to show how to extract embeddings from the model—I might even highlight this in the Basketball AI demo.

After that, we can explain how the `ReIDModel` is used by trackers like DeepSORT, with a few practical examples. From there, we can introduce the idea that in some cases, you may need to fine-tune the model. That’s where the `DataLoader` comes in—so we should explain the expected data format and link to a Google Colab showing how to use it end-to-end. Finally, we should demonstrate how to load a fine-tuned model.

Overall, we need to guide users through the full story.

trackers/core/reid/model.py
```diff
@@ -0,0 +1,477 @@
...
+            try:
+                from trackers.core.reid.callbacks import WandbCallback
+
+                callbacks.append(WandbCallback(config=config))
+            except (ImportError, AttributeError) as e:
+                logger.error(
+                    "Metric logging dependencies are not installed. "
+                    "Please install it using `pip install trackers[metrics]`."
+                )
+                raise e
```
SkalskiP, line 410: should we halt the entire training if the logger fails to load, or just display a warning instead?

SkalskiP (changes requested):

trackers/core/reid/callbacks.py
soumik12345: Here is an example:

image
image
image

docs/trackers/core/deepsort/tracker.md
soumik12345: I'm aiming to reproduce the results from the paper [Bag of Tricks and A Strong Baseline for Deep Person Re-identification](https://arxiv.org/pdf/1903.07071) and since they mention a pre-trained ResNet50 as their baseline in section 2, I just aligned with them as much as possible.

trackers/core/reid/callbacks.py
soumik12345: If we want to implement joint detection and tracking methods like [Trackformer](https://arxiv.org/pdf/2101.02702) or [FairMOT](https://arxiv.org/pdf/2004.01888), we would need training workflows other than ReID.

trackers/core/reid/model.py
onuralpszr: Imo we should just give warning with extra message like no logging and continue to train

soumik12345: IMO, when someone passes `log_to_tensorboard=True`, they're expecting the metrics to be logged to Tensorboard while training, otherwise they would have passed `False`. We should not silently skip logging since its explicitly specified by the user.

docs/trackers/core/reid/reid.md
soumik12345: Updated the header.

docs/trackers/core/reid/reid.md
soumik12345: Hi @SkalskiP, I updated the doc to act as a guide as per your feedback.

trackers/core/reid/model.py
ayulockin: As long as the `log_to_x` is not set to True as default, I agree with @soumik12345 to `raise` because it will be explicitly set by the user.

As a user, if I set True and I don't see the `wandb` url, I will be confused and in the stdout might miss to see the warning log.

ayulockin: However, if it has to fail, it should fail as soon as possible. As a user, I would not like to see error after waiting 2 mins for model loading and 3 minutes for data prep (throwing numbers but you get the point. :)).

SkalskiP: Let’s set `log_to_tensorboard` and `log_to_wandb` to `False` by default, so the user has to explicitly enable them. And as @ayulockin mentioned, let’s fail fast—raise an error early if logging is enabled but the required dependencies aren’t available.

soumik12345: They're set to `False` by default.

trackers/core/reid/callbacks.py
SkalskiP: @soumik12345 can we make it look more like [this](https://github.com/roboflow/rf-detr/blob/develop/rfdetr/util/metrics.py)?

427172819-172aee78-6b57-47f6-be6e-e2bf3305f0d5

soumik12345: Hi @SkalskiP
The plots look like this now

image

soumik12345: Updated plot:
image

docs/trackers/core/deepsort/tracker.md
SkalskiP: Oh, that looks like a really interesting paper—I'll make sure to give it a read as well.

trackers/core/reid/callbacks.py
SkalskiP: Let’s work on improving this chart a bit more. I’m thinking of two possible directions: a “maximum” and a “minimum” scenario:

- Maximum scenario: a 2x1 grid—loss on the left, accuracy on the right. We merge batch and epoch plots into a single chart. Test batch values are shown as a semi-transparent blue line without markers. Test epoch and validation epoch are displayed the same way.

- Minimum scenario: a 2x2 grid—just reorganize the charts: top left for batch triplet accuracy, top right for epoch triplet accuracy, bottom left for batch loss, and bottom right for epoch loss.

trackers/core/reid/model.py
```diff
@@ -0,0 +1,548 @@
...
+
+    @classmethod
+    def from_timm(
+        cls,
+        model_name_or_checkpoint_path: str,
+        device: Optional[str] = "auto",
+        pretrained: bool = True,
+        get_pooled_features: bool = True,
+        **kwargs,
+    ) -> "ReIDModel":
```
SkalskiP: if you add `from __future__ import annotations`, you won’t need quotes around `ReIDModel`.

trackers/core/reid/model.py
```diff
@@ -0,0 +1,548 @@
...
+                model_name_or_checkpoint_path,
+                device,
+                pretrained,
+                get_pooled_features,
+                **kwargs,
+            )
+
+    def extract_features(
+        self, frame: np.ndarray, detections: sv.Detections
+    ) -> np.ndarray:
```
SkalskiP: It occurred to me that the `BaseTrackerWithFeatures.update` API uses a different argument order. I know `extract_features` serves a different purpose, but it might be smart to keep the argument order consistent.

```python
class BaseTrackerWithFeatures(ABC):
    @abstractmethod
    def update(self, detections: sv.Detections, frame: np.ndarray) -> sv.Detections:
        pass
```

trackers/core/reid/model.py
```diff
@@ -0,0 +1,548 @@
...
+                crop = sv.crop_image(image=frame, xyxy=[*box.astype(int)])
+                tensor = self.inference_transforms(crop).unsqueeze(0).to(self.device)
+                feature = (
+                    torch.squeeze(self.backbone_model(tensor)).cpu().numpy().flatten()
+                )
+                features.append(feature)
+
+        return np.array(features)
+
+    def add_projection_layer(
```
SkalskiP: should this method be public?

trackers/core/reid/model.py
```diff
@@ -0,0 +1,548 @@
...
+            else [ToPILImage(), transforms]
+        )
+        self.model_metadata = model_metadata
+
+    @classmethod
+    def from_timm(
+        cls,
+        model_name_or_checkpoint_path: str,
+        device: Optional[str] = "auto",
+        pretrained: bool = True,
```
SkalskiP: I feel like this parameter is quite confusing—especially considering how `model_name_or_checkpoint_path` works. Users might assume that `pretrained` controls whether they’re loading a pre-trained or fine-tuned checkpoint. Is `pretrained` necessary?

trackers/core/reid/model.py
```diff
@@ -0,0 +1,548 @@
...
+            return _initialize_reid_model_from_timm(
+                cls,
+                model_name_or_checkpoint_path,
+                device,
+                pretrained,
+                get_pooled_features,
+                **kwargs,
+            )
+
+    def extract_features(
```
SkalskiP, line 159: would be cool if we could make it work with PIL and np.arrays

SkalskiP (changes requested):

trackers/core/reid/model.py
soumik12345: I would say we need the `pretrained` argument, considering that in `timm.create_model`, `pretrained=False` by default.

soumik12345:
I decided to go with the minimum scenario at https://github.com/roboflow/trackers/pull/49#discussion_r2091212071. This is how the matplotlib chart looks now 

image

trackers/core/reid/model.py
SkalskiP: Honestly I'd cat it out of our API. Always pass `True` internally. Just to keep it simple. And add it if someone complain in the future.

SkalskiP (approved):

PR #65, opened 2025-05-26
fix(trackers): bytetrack feature association

soumik12345:
# Description

This PR modifies `ByteTrackTracker` so that a track's appearance features are updated when they are successfully matched with a high-probability detection.


AlexBodner (commented):
LGTM!

PR #68, opened 2025-05-27
feat(trackers): implementation of gallery of features

soumik12345:
# Description

Add implementation of gallery of features from section 2.2 of [DeepSORT paper](https://arxiv.org/pdf/1703.07402).


rolson24 (approved):
Looks great!

Issue #59, opened 2025-05-18
Feat: Offline Tracking


Clemens-E:
Hi, first of all, thanks for your work on this project—it's a great contribution to the computer vision community!

I wanted to ask if there are any plans to support offline (batch/global) tracking algorithms in addition to the current online trackers. While real-time, online tracking is essential for many use cases (especially with YOLO and similar models), offline tracking can be extremely valuable for post-processing tasks where speed isn’t as critical, such as video analytics, privacy blurring, or high-accuracy benchmarking.

Thank you for all the hard work that went into this

SkalskiP:
Thanks so much for the kind words—we really appreciate the support!

Yes, adding offline (batch/global) trackers is definitely on our roadmap. Our next 1–2 months are pretty tightly scheduled, but we’re planning to start working on offline tracking support after that. Of course, if we get help from open-source contributors in the meantime, we could begin integrating offline trackers even sooner.

Are there any specific offline tracking algorithms you'd recommend or be particularly interested in seeing added? We'd love to hear your suggestions!

SkalskiP:
I’m closing this issue as resolved.

@Clemens-E, feel free to reopen it if you have any follow-up questions.

PR #89, opened 2025-06-17
Feature: Offline Tracker (KSP Tracker)

Ashp116:
# Description

This PR introduces a new **offline Multiple Object Tracker** (MOT) based on the **K-Shortest Paths (KSP)** optimization algorithm. This tracker is based on the method proposed in the paper: [**"Multiple Object Tracking using K-Shortest Paths Optimization" by Jérôme Berclaz et al**](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf).  The tracker models detections across frames as a **directed acyclic graph** and solves for globally optimal object trajectories by computing disjoint shortest paths.

The inspiration for this PR: #59 

### List any dependencies that are required for this change.
Yes, this PR needs [`networkx`](https://networkx.org/documentation/stable/) to build a directed acyclic graph.

## Type of change

-   [➖] This change requires a documentation update: _maybe? I'm not sure_

## How has this change been tested? Please provide a testcase or example of how you tested the change?

This PR can be tested using the Colab notebook provided [here](https://colab.research.google.com/drive/12EQygTxynXRNrpQPKg0d7il2RRUB2Mr6?usp=sharing). The notebook can be used to test KSP Tracker. Currently, the notebook is using one of the [MOT17 Challenge Datasets](https://motchallenge.net/data/MOT17/), specifically MOT17-02-FRCNN.

N/A

## Docs

-   [❌] Docs updated? What were the changes? _Not yet_


trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
+from dataclasses import dataclass
+from typing import Any, Dict, List, Union
+
+import networkx as nx
```
soumik12345: Please add `networkx` as a optional dependency under `ksptracker`. Simply run the following command and it should update `pyproject.toml` and `uv.lock`

```bash
uv add networkx --optional ksptracker
```

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+        Args:
+            detections (sv.Detections): New detections for current frame
+
+        Returns:
+            sv.Detections: Input detections (unmodified)
+        """
+        self.detection_buffer.append(detections)
+        return detections
+
+    def _calc_iou(
```
soumik12345: The [`box_iou_batch`](https://supervision.roboflow.com/latest/detection/utils/) function from supervision seems like a drop-in replacement from this.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        paths: List[List[TrackNode]] = []
+        for path in nx.shortest_simple_paths(graph, "source", "sink", weight="weight"):
+            if len(paths) >= self.max_paths:
+                break
+            # Remove source and sink nodes from path
+            paths.append(path[1:-1])
+        return paths
+
+    def process_tracks(self) -> sv.Detections:
```
soumik12345: IMO, `process_tracks` should return a `list[sv.Detections]` where a list of tracklets are mapped to the corresponding frame of the video.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+
+        Returns:
+            sv.Detections: Detections with assigned track IDs
+        """
+        graph = self._build_graph(self.detection_buffer)
+        paths = self.ksp(graph)
+
+        # Assign track IDs
+        assignments = {}
+        for track_id, path in enumerate(paths, start=1):
```
soumik12345: Since this loop takes a significant amount of time to execute, please wrap the `enumerate` with a `tqdm` progress bar. For example, like the following:

```python
from tqdm.auto import tqdm

for track_id, path in tqdm(enumerate(paths, start=1), total=len(paths), desc="Processing tracks"):
```

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+            for node in path:
+                assignments[(node.frame_id, node.detection_id)] = track_id
+
+        return
```
soumik12345: As I mentioned in my previous comment, please make sure `process_tracks` returns a `list[sv.Detections]` where a list of tracklets are mapped to the corresponding frame of the video. I think it misses the `_update_detections_with_tracks` call to apply the track IDs to the detections and return the final result.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+
+import networkx as nx
+import numpy as np
+import supervision as sv
+
+from trackers.core.base import BaseTracker
+
+
+@dataclass(frozen=True)
+class TrackNode:
```
soumik12345: It seems like this implementation uses the detections themselves as graph nodes, while the paper proposes a graph where nodes represent a grid of discrete physical locations at every point in time which allows the tracker to interpolate trajectories through areas with missing detections.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+        """Find K-shortest paths in the graph.
+
+        Args:
+            graph (nx.DiGraph): Directed graph of detection nodes
+
+        Returns:
+            List[List[TrackNode]]: List of paths, each path is list of TrackNodes
+        """
+        paths: List[List[TrackNode]] = []
+        for path in nx.shortest_simple_paths(graph, "source", "sink", weight="weight"):
```
soumik12345: `nx.shortest_simple_paths` generates all paths without repeated nodes between a source and target in order of increasing length, but that does not necessarily guarantee node-disjoint paths. This means that a single detection could be assigned to multiple tracks.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,277 @@
...
+            sv.Detections: Detections with assigned track IDs
+        """
+        graph = self._build_graph(self.detection_buffer)
+        paths = self.ksp(graph)
+
+        # Assign track IDs
+        assignments = {}
+        for track_id, path in enumerate(paths, start=1):
+            for node in path:
+                assignments[(node.frame_id, node.detection_id)] = track_id
```
soumik12345: This simply overwrites the `track_id`, meaning only the last path found containing that detection gets the assignment. This doesn't seem like an accurate representation of the disjoint-path formulation described in the paper.

soumik12345 (changes requested):
Hi @Ashp116, thanks for the PR!

I have mentioned some feedback regarding the accuracy of the implementation as per the paper and also regarding how the API for using a offline tracker should be like.

Ashp116:
Hi @soumik12345, thank you for reviewing the PR!

I’ve applied the suggested changes—they’ve significantly improved the tracker. I also appreciate you pointing out that the KSP solver doesn't guarantee node-disjoint paths. After the update, `process_tracks()` now runs instantaneously.

I've also updated the [collab notebook](https://colab.research.google.com/drive/12EQygTxynXRNrpQPKg0d7il2RRUB2Mr6?usp=sharing) accordingly.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,252 @@
...
+            for node in node_dict[i]:
+                for node_next in node_dict[i + 1]:
+                    dist = np.linalg.norm(
+                        np.array(node.position) - np.array(node_next.position)
+                    )
+                    if dist <= 2:
+                        G.add_edge(
+                            node_next,
+                            node,
+                            weight=max(dist - node_next.confidence, 0.001),
```
soumik12345: I am a little unsure as to how accurate this edge-weighing strategy is. This is what section 3.3 of the paper states:

> A directed edge $e_{i, j}^{t}$ from location $i$ at time $t$ to location $j$ at time $t+1$ is assigned the cost value $$c(e_{i, j}^{t}) = -log(\frac{\rho_{i}^{t}}{1 - \rho_{i}^{t}})$$
The cost value of the edges emanating from the source node is set to zero to allow objects to appear at any entrance position and at any time instant at no cost.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,252 @@
...
+                )
+                if len(path) < 2:
+                    break
+                paths.append(path[1:-1])
+                for node in path[1:-1]:
+                    G_copy.remove_node(node)
+            except nx.NetworkXNoPath:
+                break
+
+        return paths
```
soumik12345: This version of KSP doesn't seem to accurately match the algorithm described in Appendix B of the paper. I'm not entirely sure if this greedy search is guaranteed to find the globally optimal set of `k` paths.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,252 @@
...
+        Initialize KSP tracker with config parameters.
+
+        Args:
+            grid_size (int): Pixel size of each grid cell
+            max_gap (int): Max frames between connected detections
+            min_confidence (float): Min detection confidence
+            max_distance (float): Max allowed dissimilarity
+        """
+        self.grid_size = grid_size
+        self.max_gap = max_gap
```
soumik12345: `self.max_gap` doesn't seem to be used in this implementation. According to the paper, a key strength of KSP Tracker is the ability to bridge gaps caused by false negatives (missed detections).

soumik12345 (changes requested):
Hi @Ashp116, thanks for addressing my feedback in the initial round of the review.
I have noted a few more differences in the implementation compared to the paper.
Also, I have a quick suggestion on the colab notebook: can you please update it to plot the tracker IDs using `sv.process_video` as shown in [this colab notebook](https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/how-to-track-objects-with-sort-tracker.ipynb)?

soumik12345:
Hi @Ashp116, I have some notes regarding how you can test your implementation of `KSPTracker`...

We don't yet have an end-to-end quantitative evaluation system that can evaluate MOT algorithms (they are being worked on as part of https://github.com/roboflow/trackers/pull/29). Hence, besides carefully checking the code for implementation correctness for the paper, we need to "vibe check" the results by plotting and observing the tracking IDs. I created a [modified version of your colab](https://colab.research.google.com/drive/1kr94TAenlp8Abdr7C2xO0orO_XNe5FvR?usp=sharing) that shows us the tracking IDs (or rather the lack of it) in the result. You can use the code in this modified notebook to test your implementation as you keep working on the PR.

I eagerly look forward to seeing this PR evolve into a valuable contribution!

Ashp116:
Hi @soumik12345, 

For my implementation of the tracker, I am currently testing it with a new dataset, one that the paper uses (PES 2009). I have also migrated to using `sv.process_video`. I have created a new branch under my fork (not yet merged to this PR) where I matched the paper's implementation as accurately as possible. However, I am encountering an issue where only 2 or 3 track paths were created from the KSP algorithm. I am a bit stuck on how to move forward from this. I am currently trying to debug the graph by creating Matplotlib plots. Here is the [branch](https://github.com/Ashp116/trackers/tree/fix/ksp-disjoints) I am currently working on.

By the way, the modified version of the collab notebook seems to navigate to a different PR.

soumik12345:
> By the way, the modified version of the collab notebook seems to navigate to a different PR.

Can you please clarify what is the source branch on your fork for this PR? The PR shows me `Ashp116:feature/offline-tracker-ksp`.

Ashp116:
> Can you please clarify what is the source branch on your fork for this PR? The PR shows me

The link you provided for the "modified version of your collab" doesn't direct me to a collab notebook, but instead to a different PR. Here is a video on what I see when I click the link. https://gyazo.com/e3ea6d8796f2dc3510691ce5d4cddfe6.mp4

soumik12345:
> The link you provided for the "modified version of your collab" doesn't direct me to a collab notebook, but instead to a different PR. Here is a video on what I see when I click the link. https://gyazo.com/e3ea6d8796f2dc3510691ce5d4cddfe6.mp4

My bad, here's the modified colab: https://colab.research.google.com/drive/1kr94TAenlp8Abdr7C2xO0orO_XNe5FvR?usp=sharing

Ashp116:
Hi @soumik12345,

I have a question regarding the process_video function and its respective annotators. I can generate some tracks from the PETS 2009 dataset, but the detection IDs jump around. I have attached pictures and pasted my debug data to show the discrepancy in the detection IDs with TrackIDs. The "ID: X" is the track ID. From one frame to another, the detection ID changes, which keeps the current track ID unchanged, giving these weird jumps.

If you notice the `Detection 0` and `Detection 1` in Frame 3 vs Frame 4, you can say that these 2 detection IDs swapped (because a new object was detected). 

Now, my question is whether this is a problem with process_video, where it doesn't properly correlate the detection's track with its respective annotator, or is this an issue with the way I am assigning track_ids to each frame? I believe it's the latter, but I want to double-check.

```
[Frame 3] Total Detections: 6 | Tracked: 3
  ✅ Track 1 assigned to Detection 0: BBox=[     273.08      216.34      307.06      302.17], Conf=0.84
  ✅ Track 2 assigned to Detection 1: BBox=[     618.01       239.5      649.46      323.73], Conf=0.82
  ✅ Track 3 assigned to Detection 2: BBox=[     493.91      160.82      522.57      236.24], Conf=0.80
  ⛔ Untracked Detection 3: BBox=[     734.22      50.179       767.6      94.558], Conf=0.73
  ⛔ Untracked Detection 4: BBox=[      652.3      44.005      721.93      99.784], Conf=0.63
  ⛔ Untracked Detection 5: BBox=[     29.699      111.64      49.206      141.46], Conf=0.33
[Frame 4] Total Detections: 7 | Tracked: 3
  ✅ Track 1 assigned to Detection 0: BBox=[     598.74      241.02      647.35      324.48], Conf=0.86
  ✅ Track 2 assigned to Detection 1: BBox=[     280.19      214.17      315.16         301], Conf=0.82
  ✅ Track 3 assigned to Detection 2: BBox=[     493.25      161.63      518.34      237.37], Conf=0.79
  ⛔ Untracked Detection 3: BBox=[     734.29      50.745      767.55      94.547], Conf=0.71
  ⛔ Untracked Detection 4: BBox=[     652.13      44.389      722.24      100.73], Conf=0.64
  ⛔ Untracked Detection 5: BBox=[      651.7      44.156      722.31      100.27], Conf=0.28
  ⛔ Untracked Detection 6: BBox=[     29.482      111.75      49.205      141.56], Conf=0.27
```

Frame 3:
image 
Frame 4:
image

soumik12345:
Hi @Ashp116, have you tried annotating the frames with the corresponding IDs without `process_video` and reproducing the same effect? If the ID swaps are recorded without `process_video`, then we can conclude that the issue lies with `KSPTracker`.

Ashp116:
Hi @soumik12345,

I explored alternative approaches for building and running KSP-based tracking. The paper I initially referenced did not clearly describe how they constructed the occupancy map, making it difficult to replicate their method directly. In my case, the only inputs available to the `KSPTracker` are per-frame detections, so following their approach would have required building the occupancy map myself — effectively estimating posterior probabilities — which added extra complexity.

While that approach offered some insight, it struggled with overlapping detections. I also came across another [paper](https://arxiv.org/pdf/1808.01562) that proposed training a model using a multi-layer perceptron to construct the tracking graph. I didn’t pursue that route, but one helpful takeaway from the paper was how other MOT methods define detection affinities:

> There are various ways to define the affinity model in stage I, like bounding box intersection-over-union (IoU), spatial-temporal distance, appearance similarity, etc.

I ended up combining a few of these affinity metrics to define edge weights in my tracker. The edge cost between two detections is computed as:

$c_{u \to v}$ = $\alpha$ * (1 - IoU) + $\beta$ * Distance + $\gamma$ * SizePenalty + $\delta$ * (1 - min(Conf_a, Conf_b))


Where:
- `IoU` is the intersection-over-union between two bounding boxes  
- `Distance` is the Euclidean distance between their centers  
- `SizePenalty` is a log penalty based on the area ratio  
- `Conf_a`, `Conf_b` are the detection confidences  
- $\alpha$,  $\beta$, $\gamma$,  $\delta$ are tunable hyperparameters

This method has worked quite well. After constructing the base graph, I used KSP to extract the top-k paths. One issue I encountered was that a single detection could be assigned to multiple trackers in the same frame. To resolve this, I selected the tracker with the least positional change between frames *t* and *t + 1*, which helped maintain temporal consistency.

I tested the tracker on the PETS 2009 dataset and attached a sample video output, along with a [Colab notebook](https://colab.research.google.com/drive/1uxIOC6U6IrUhpd7A8-ktUZzLXsXaWrM2?usp=sharing) demonstrating the implementation.

Let me know what you think!

https://github.com/user-attachments/assets/e3b0a253-e2f7-45dd-b2c5-a70a9a8b5a75

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+        return isinstance(other, TrackNode) and (self.frame_id, self.det_idx) == (
+            other.frame_id,
+            other.det_idx,
+        )
+
+    def __str__(self):
+        return f"{self.frame_id}:{self.det_idx}@{self.position}"
+
+
+class KSP_Solver:
```
soumik12345: Please rename this class to `KSPSolver`.

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+        center_dist = np.linalg.norm(self._get_center(a) - self._get_center(b))
+        iou_penalty = 1 - self._iou(a, b)
+
+        area_a = (a[2] - a[0]) * (a[3] - a[1])
+        area_b = (b[2] - b[0]) * (b[3] - b[1])
+        size_penalty = np.log(
+            (max(area_a, area_b) / (min(area_a, area_b) + 1e-6)) + 1e-6
+        )
+
+        conf_penalty = 1 - min(conf_a, conf_b)
```
soumik12345: Have you experimented with how the results look if the confidence penalty is implemented according to Eqn 12 from Section 3.3, [Berclaz et al.](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf)?

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+        Args:
+            bbox (np.ndarray): Bounding box coordinates.
+
+        Returns:
+            np.ndarray: Center coordinates.
+        """
+        x1, y1, x2, y2 = bbox
+        return np.array([(x1 + x2) / 2, (y1 + y2) / 2])
+
+    def _iou(self, a, b):
```
soumik12345: The [`box_iou_batch`](https://supervision.roboflow.com/latest/detection/utils/) function from supervision seems like a drop-in replacement for this.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,125 @@
...
+                    dy = node.position[1] - next_node.position[1]
+                    displacement = dx * dx + dy * dy  # squared distance
+                else:
+                    displacement = 0  # last node in path, no penalty
+
+                if displacement < min_displacement:
+                    min_displacement = displacement
+                    selected_tracker = tracker_id
+
+            node_to_tracker[node] = selected_tracker
```
soumik12345: Is this how the node-disjoint constraint is enforced?

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+                    G.add_edge(node_a, node_b, weight=cost)
+
+        for node in node_frames[0]:
+            G.add_edge(self.source, node, weight=0.0)
+        for node in node_frames[-1]:
+            G.add_edge(node, self.sink, weight=0.0)
+
+        self.graph = G
+
+    def solve(self, k: Optional[int] = None) -> List[List[TrackNode]]:
```
soumik12345: If I understand this logic correctly, it still doesn't seem to guarantee node-disjoint paths.

Currently, it iteratively finds the shortest path using Dijkstra's algorithm; after a path is found, it increases the cost of the edges in that path.

This makes it less likely for future paths to use the same edges, which indirectly makes it less likely for them to use the same nodes. However, it does not make it impossible. A single `TrackNode` can still appear in multiple paths if the alternative routes are significantly more costly, unless its constrained in `KSPTracker.assign_tracker_ids_from_paths()`.

It would be really helpful if you can shed some light on this comment, and help me understand the logic.

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+            min(a[3], b[3]),
+        )
+        inter = max(0, x2 - x1) * max(0, y2 - y1)
+        area_a = (a[2] - a[0]) * (a[3] - a[1])
+        area_b = (b[2] - b[0]) * (b[3] - b[1])
+        return inter / (area_a + area_b - inter + 1e-6)
+
+    def _edge_cost(
+        self, a, b, conf_a, conf_b, iou_w=0.5, dist_w=0.3, size_w=0.1, conf_w=0.1
+    ):
```
soumik12345: Please add type hints for this function.

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+            except nx.NetworkXNoPath:
+                break
+
+            if path[1:-1] in paths:
+                break
+
+            paths.append(path[1:-1])
+
+            for u, v in zip(path[:-1], path[1:]):
+                edge_reuse[(u, v)] += 1
```
soumik12345: Please put nested `tqdm` progress bars in these for loops with appropriate descriptions.

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,230 @@
...
+            min(a[2], b[2]),
+            min(a[3], b[3]),
+        )
+        inter = max(0, x2 - x1) * max(0, y2 - y1)
+        area_a = (a[2] - a[0]) * (a[3] - a[1])
+        area_b = (b[2] - b[0]) * (b[3] - b[1])
+        return inter / (area_a + area_b - inter + 1e-6)
+
+    def _edge_cost(
+        self, a, b, conf_a, conf_b, iou_w=0.5, dist_w=0.3, size_w=0.1, conf_w=0.1
```
soumik12345: Please make the coefficients for computing edge cost configurable from `KSPTracker`.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,125 @@
...
+from trackers.core.base import BaseTracker
+from trackers.core.ksptracker.solver import KSP_Solver, TrackNode
+
+
+class KSPTracker(BaseTracker):
+    """
+    Offline tracker using K-Shortest Paths (KSP) algorithm.
+    """
+
+    def __init__(self) -> None:
```
soumik12345: The hyperparameters `iou_w`, `dist_w`, `size_w`, and `conf_w` should be configurable while initializing `KSPTracker`

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,125 @@
...
+                xyxy=xyxy,
+                confidence=confidence,
+                class_id=class_id,
+                tracker_id=tracker_id,
+            )
+            frame_to_detections.append(detections)
+
+        return frame_to_detections
+
+    def process_tracks(self, num_of_tracks=None) -> List[sv.Detections]:
```
soumik12345: We should have a single-point-of-entry function for `KSPTracker` that accepts the path to a video, the path to a directory of frames as image files, or a list of frames and returns a list of `sv.Detections` mapping to corresponding frame index. The logic of the function could be something like the following:

```python
def track(self, source_path, num_of_tracks: Optional[int] = None) -> list[sv.Detections]:
    frames_generator = sv.get_video_frames_generator(source_path=source_path)
    video_info = sv.VideoInfo.from_video_path(video_path=source_path)

    for frame in tqdm(frames_generator, total=video_info.total_frames):
        result = model(frame, verbose=False)[0]
        detections = sv.Detections.from_ultralytics(result)
        tracker.update(detections)
    
    paths = self._solver.solve(num_of_tracks)
    if not paths:
        return []
    return self.assign_tracker_ids_from_paths(paths)
```

soumik12345 (changes requested):
Hi @Ashp116, thanks for the work on this PR so far. Here's my overall feedback:

✅ I think your approach of building a graph where nodes represent only the actual detections is an acceptable approximation of the probabilistic occupancy maps recommended in the [paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf).

✅ I like the modified edge cost function using the affinity metrics, however I still think that we should try to align the confidence component of the cost function to align with the [paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf).

❓ I'm a little unsure regarding how well the current implementation can guarantee node-disjoint shortest paths. My understanding from the paper is that the authors use Suurballe's algorithm (refer to the 1st paragraph of Appendix B in the [paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf)). Can you share your thoughts on this?

soumik12345:
Hi @Ashp116, I tried using `KSPTracker` on a different video, and when I captured the detections at `imgsz=1280`, the tracker seemed to fail to track most of the targets. Do you have a hypothesis as to why this might be happening? Here's the [colab](https://colab.research.google.com/drive/1ygslmbJeJFKug-ID3nZtFSHvTB_qqtvx?usp=sharing) that I used.


https://github.com/user-attachments/assets/46e3e42f-1f1e-438b-aff3-f097eafc6346

Ashp116:
Hi @soumik12345,

Thank you for reviewing the PR. I will respond to the question in the code review as soon as possible. However, I was surprised by the result you obtained. When I ran the same input, I was able to generate significantly more tracks. I'll attach a video showing what I generated.

I believe this discrepancy is due to the `process_tracks` function not being given enough time to generate all the tracks.

https://github.com/user-attachments/assets/723a797f-4f8a-4561-93af-9cdd10764a49

trackers/core/ksptracker/solver.py
Ashp116: Yes, this is correct. `solve` doesn't guarantee node-disjoints, but the `assign_tracker_ids_from_paths` does. The reason I went with this approach is produces fewer tracer jumps/switches. I did test where I made this function only get the disjoint-paths from source to sink, but it still produced tracer switches. In addition, it didn't show some objects in certain frames. With reference to the video that you used in the collab, it only got 31/47 paths. Do note that 47 is the max number of detections in a frame across the entire video. 

My reasoning in going with this logic is, although some objects might not be detected at the start of the video, but show up after some time, the path that the tracker takes will be through nodes that probably share other trackers. This frame of logic allows us to still create trackers for those objects. There is still the problem of disjoints, but this is taken care of with `assign_tracker_ids_from_paths`.

trackers/core/ksptracker/tracker.py
Ashp116: Yes, this part of the code enforces the node-disjointness for a given tracker path.

trackers/core/ksptracker/solver.py
Ashp116: Yes, I spent a lot of time experimenting with the edge cost that the paper proposes. From [from all the commits after the previous code review till this commit](https://github.com/Ashp116/trackers/blob/ccd77af3f0587842b7d8cf60d6eca403e45ea33a/trackers/core/ksptracker/tracker.py#L715), I spent a lot of time trying to debug. I would like to give you a heads up that this part of the code is messy. I have a lot of other functions that I used to try to visualize the graph. I did build an occupancy map, although I referred to it as a probability map in my code.

soumik12345:
> I believe this discrepancy is due to the `process_tracks` function not being given enough time to generate all the tracks.

Hi @Ashp116,
Do you mean setting the `num_of_tracks` parameter in `process_tracks` to a higher value?

Ashp116:
> Do you mean setting the num_of_tracks parameter in process_tracks to a higher value?

No, I mean like the execution time. 

Edit: My response is a bit ambiguous. What I mean is the time it takes for all the tracks/paths to be computed; in total, 47 paths for the given video. I will push my changes with tqdm, try it out and let me know.

trackers/core/ksptracker/solver.py
Ashp116: I switched to `box_iou_batch`, but it seems to be significantly slower than the `solve` algorithm. Here are my tqdm stats and the code I used; It's about 2x slower.


**box_iou_batch**
```
Running detection and buffering...
100%|████████████████████████████████████████████████████████████████████████████████| 341/341 [01:05<00:00,  5.24it/s]
Extracting k-shortest paths:  23%|████████████▍                                        | 11/47 [00:41<02:21,  3.93s/it]
```
__code__: `return sv.box_iou_batch(np.array([a]), np.array([b]))[0]`

*My iou func*

```
Running detection and buffering...
100%|████████████████████████████████████████████████████████████████████████████████| 341/341 [01:04<00:00,  5.28it/s]
Extracting k-shortest paths:  23%|████████████▍                                        | 11/47 [00:24<01:27,  2.42s/it]
```
__code__:  same function submitted for the code review

Ashp116:
Hi @soumik12345,

What should be the expected behavior of a tracker? I'm trying to ensure that each tracker consistently follows a single object with minimal external switching. I've fine-tuned the parameters and achieved slightly better results than in the previous video. However, I'd appreciate your input on what the ideal behavior of a tracker should look like.



https://github.com/user-attachments/assets/af04b60e-a454-498e-97d3-1ea6f5fa9eb0

soumik12345:
> What should be the expected behavior of a tracker? I'm trying to ensure that each tracker consistently follows a single object with minimal external switching

Following a single object with minimal external switching is the core objective of a tracking algorithm, but there are a few more things that an ideal tracker should be robust against, such as partial or total occlusions, motion blurs, missed detections, and suppressing short-lived false positives and detector noise.

If you're asking for a single quantitative metric, there are several metrics for association accuracy, such as MOTA, MOTP, HOTA, etc. We are currently working on tools for quantitative benchmarking of trackers at https://github.com/roboflow/trackers/pull/29

From a qualitative perspective, I find the aforementioned result acceptable, since it confirms that the tracker can associate detections across frames; quantitative metrics can further measure how robust they are, especially under challenging scenarios.

trackers/core/ksptracker/solver.py
soumik12345: @SkalskiP do you have any thoughts here?

Ashp116:
Hi @soumik12345,

I made some changes around tracker tuning and introduced a unified entry point via `process_tracks`.

To start with, `process_tracks` now takes three optional parameters:

* `source_path`: Accepts either a video file or a folder of JPEGs, which are then passed into `KSPTracker`.
* `get_model_detections`: A callback function that allows flexible ways to retrieve `sv.Detections`.
* `num_of_tracks`: Optional override for setting the number of tracks manually.

On the tracker tuning side—while running `yolov8n` locally, I noticed a significant mismatch in results compared to what you showed in one of your earlier comments. Interestingly, switching to `yolo11x` brought the outputs much closer to your results. Specifically, it produced fewer tracks than detections, which initially seemed like a bug.

After digging into it, I found the root cause: the reuse of nodes was too cheap in the cost graph, so the tracker preferred to follow those paths rather than create new ones. This led to duplicate tracks. I addressed this by increasing the `path_overlap_penalty` (previously called `base_penalty`), which discourages reusing the same nodes across paths. This adjustment improved the tracking consistency noticeably.

As a result, I’ve exposed all relevant hyperparameters—edge weights and overlap penalties—as configurable values. This should make fine-tuning easier down the line.

Here is the [collab notebook](https://colab.research.google.com/drive/1eAJ7HzxWpWtY3NAwMtEYhJXT82lc2VZQ?usp=sharing) to test around.

I’ll attach a video comparison showing the before/after effect and the corresponding parameter values.

Let me know what you think!

### Videos

**Video showcasing fewer tracks for video detections**
__config__
```
path_overlap_penalty= 10 
iou_weight= 1
dist_weight= 0.5 
size_weight= 0.2
conf_weight= 0.15
```

https://github.com/user-attachments/assets/a3b8999d-6460-4595-8079-2210f24aafc9

---

**Video showcasing maximizing tracks for video detections**
__config__
```
path_overlap_penalty= 40
iou_weight= 0.9
dist_weight= 0.1
size_weight= 0.1
conf_weight= 0.1
```
https://github.com/user-attachments/assets/ff91dbb2-37be-4f18-be8a-fde60ee3254d

Ashp116:
>❓ I'm a little unsure regarding how well the current implementation can guarantee node-disjoint shortest paths. My understanding from the paper is that the authors use Suurballe's algorithm (refer to the 1st paragraph of Appendix B in the [paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf)). Can you share your thoughts on this?

Apologies for not addressing this earlier — I touched on it in another comment, but I’ll clarify it here.

The goal of the KSP approach I implemented is to generate paths that are **not guaranteed to be disjoint**. This is intentional. The idea is to allow detections that may occur in distant frames to still be connected and tracked. For example, if there are 10 detections early on but 30 appear at frame 100, a strict disjoint constraint would discard many of them since there would be no feasible path from source to sink.

Additionally, while the original paper ([Berclaz et al.](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf)) mentions Suurballe's algorithm, my understanding is that Suurballe is specifically designed to find **two disjoint paths**, not *k* paths. That seems limiting in this context. The paper compares the cost difference between $p_l$ and $p_{l-1}$, which makes sense in the paper's context.

I experimented with other algorithms, including Yen’s algorithm. However, Yen’s method struggles with performance on large graphs — it becomes impractical at scale. The approach I implemented is significantly faster and yields reasonable results, especially when tracking across many frames.

trackers/core/ksptracker/solver.py
soumik12345: My personal thought on this is that we should stick to using `box_iou_batch` here to maintain parity with the design of rest of the trackers. If you have an idea of how to speed it up, let's discuss it on [supervision](https://github.com/roboflow/supervision).

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,310 @@
...
+    top_left = np.maximum(boxes_true[:, None, :2], boxes_detection[:, :2])
+    bottom_right = np.minimum(boxes_true[:, None, 2:], boxes_detection[:, 2:])
+
+    wh = np.clip(bottom_right - top_left, a_min=0, a_max=None)
+    area_inter = wh[:, :, 0] * wh[:, :, 1]
+
+    ious = area_inter / (area_true[:, None] + area_detection - area_inter)
+
+    ious = np.nan_to_num(ious)
+    return ious
```
soumik12345: What's the purpose of this code?

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,310 @@
...
+        if path_overlap_penalty is not None:
+            self.path_overlap_penalty = path_overlap_penalty
+        if iou_weight is not None:
+            self.weights["iou"] = iou_weight
+        if dist_weight is not None:
+            self.weights["dist"] = dist_weight
+        if size_weight is not None:
+            self.weights["size"] = size_weight
+        if conf_weight is not None:
+            self.weights["conf"] = conf_weight
```
soumik12345: I think that the initialization can be moved to the constructor of `KSPSolver`.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,220 @@
...
+    Offline tracker using K-Shortest Paths (KSP) algorithm.
+    """
+
+    def __init__(
+        self,
+        path_overlap_penalty: Optional[int] = None,
+        iou_weight: Optional[int] = None,
+        dist_weight: Optional[int] = None,
+        size_weight: Optional[int] = None,
+        conf_weight: Optional[int] = None,
```
soumik12345: Please document the purpose and effects of these parameters in the docstring.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,220 @@
...
+            size_weight (Optional[float]): Weight for size component.
+            conf_weight (Optional[float]): Weight for confidence component.
+        """
+        self._solver.append_config(
+            path_overlap_penalty=path_overlap_penalty,
+            iou_weight=iou_weight,
+            dist_weight=dist_weight,
+            size_weight=size_weight,
+            conf_weight=conf_weight,
+        )
```
soumik12345: Refer to my earlier comment on the same wrt `KSPSolver`.

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,220 @@
...
+import cv2
+import numpy as np
+import supervision as sv
+from tqdm import tqdm
+
+from trackers.core.base import BaseTracker
+from trackers.core.ksptracker.solver import KSPSolver, TrackNode
+
+
+class KSPTracker(BaseTracker):
```
soumik12345: Going forward, we would like to make offline trackers first class member of the library. Keeping that in mind, let's make `KSPTracker` inherit from a different base class `BaseOfflineTracker` that needs a single `track` function to be overloaded (similar to `update` for `BaseTracker`).

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,220 @@
...
+                xyxy=xyxy,
+                confidence=confidence,
+                class_id=class_id,
+                tracker_id=tracker_id,
+            )
+            frame_to_detections.append(detections)
+
+        return frame_to_detections
+
+    def process_tracks(
```
soumik12345: Let's rename this function to `track` (refer to my earlier comment on modifying the base class of `KSPTracker`).

trackers/core/ksptracker/solver.py
```diff
@@ -0,0 +1,310 @@
+from collections import defaultdict
+from dataclasses import dataclass
+from typing import Any, List, Optional, Tuple
+
+import networkx as nx
+import numpy as np
+import supervision as sv
+from tqdm import tqdm
```
soumik12345:
```suggestion
from tqdm.auto import tqdm
```

trackers/core/ksptracker/tracker.py
```diff
@@ -0,0 +1,220 @@
+import os
+from collections import defaultdict
+from typing import Any, Callable, List, Optional
+
+import cv2
+import numpy as np
+import supervision as sv
+from tqdm import tqdm
```
soumik12345:
```suggestion
from tqdm.auto import tqdm
```

soumik12345 (changes requested):
Hi @Ashp116, great work on the PR so far!
I think we are really close to getting it merged

✅ I think we can settle on implementation correctness and agree that the deviations that you have taken from the [original paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf) is acceptable.

👉 I've left some additional feedback on the implementaion and the API.

📕 Also, I think its time to start adding a dedicated documentation page for `KSPTracker` that documents how the tracker functions, how to use it, and how it differs in implementation from the [original paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf).

Also, I think its time @SkalskiP takes a look at the PR.

trackers/core/ksptracker/solver.py
Ashp116: Oh woops, I was trying to figure out why `box_iou_batch` was slower, and how to improve the speed.

Ashp116:
Hi @soumik12345,

Thank you for reviewing the PR, and also thank you for your kind words. I’m really happy this will be merged soon.

I have made the requested changes and also added documentation. Regarding the `box_iou_batch` function, I find it to be fast when computing IoUs for large batches of bounding boxes. However, in situations with only a few bounding boxes to compare, it takes more time than the original function. Of course, we can discuss this further in the Supervision repo, but I wanted to mention that having a dedicated IoU function optimized for single bounding box comparisons might be useful.

> how it differs in implementation from the original paper.

Just curious — how in-depth should I go when explaining the implementation differences? Should I keep it high-level, or go into the full algorithm and cost differences compared to the paper? Also, should I explicitly mention this [paper](https://arxiv.org/pdf/1808.01562) in the documentation when explaining the differences?

Also, I was experimenting with the tracker docs UI style. I was wondering if the new tracker docs style (showcased in this [video](https://i.gyazo.com/962548d18fc955c4151e20b29a934933.mp4)) might be something to consider changing? It fits well with the rest of the docs (e.g., Supervision, Inference, etc.). If such a change is appropriate, should I push it to this PR or create a separate one?

Thanks again!

docs/trackers/core/ksp/tracker.md
```diff
@@ -0,0 +1,185 @@
+---
+comments: true
+---
+
+# KSP
+
+[![IEEE](https://img.shields.io/badge/IEEE-10.1109/TPAMI.2011.21-blue.svg)](https://doi.org/10.1109/TPAMI.2011.21)
+[![PDF (Unofficial)](https://img.shields.io/badge/PDF-Stanford--Preprint-red.svg)](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf)
+[![colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/roboflow-ai/notebooks/blob/main/notebooks/how-to-track-objects-with-sort-tracker.ipynb)
```
soumik12345, line 10: Please fix this colab link. Can you please raise a corresponding PR on [roboflow/notebooks](https://github.com/roboflow/notebooks)?

trackers/core/ksp/tracker.py
```diff
@@ -0,0 +1,219 @@
...
+                tracker_id=tracker_id,
+            )
+            frame_to_detections.append(detections)
+
+        return frame_to_detections
+
+    def track(
+        self,
+        source_path: str,
+        get_model_detections: Callable[[np.ndarray], sv.Detections],
```
soumik12345, line 228: This should just be a list of `sv.Detections`, not a function.

trackers/core/ksp/tracker.py
```diff
@@ -0,0 +1,219 @@
...
+        if not source_path:
+            raise ValueError(
+                "`source_path` must be a string path to a directory or an .mp4 file."
+            )
+        if get_model_detections is None:
+            raise TypeError(
+                "`get_model_detections` must be a callable that returns an "
+                "instance of `sv.Detections`."
+            )
+        if source_path.lower().endswith(".mp4"):
```
soumik12345: I think you should just check if the `source_path` is a file or not, if its a file, lets assume it to be a video file, otherwise check if its a directory of images or not.

trackers/core/ksp/tracker.py
```diff
@@ -0,0 +1,219 @@
...
+                class_id=class_id,
+                tracker_id=tracker_id,
+            )
+            frame_to_detections.append(detections)
+
+        return frame_to_detections
+
+    def track(
+        self,
+        source_path: str,
```
soumik12345: Let's rename this to be just `source` and make sure that it can also handle a list of frames (`PIL.Image.Image`) besides being a path to a video or frames.

trackers/core/ksp/tracker.py
```diff
@@ -0,0 +1,219 @@
...
+    Offline tracker using K-Shortest Paths (KSP) algorithm.
+    """
+
+    def __init__(
+        self,
+        path_overlap_penalty: Optional[int] = None,
+        iou_weight: Optional[int] = None,
+        dist_weight: Optional[int] = None,
+        size_weight: Optional[int] = None,
+        conf_weight: Optional[int] = None,
```
soumik12345: I think its best to give them some optimal default value instead of making them optional parameters.

trackers/core/ksp/solver.py
```diff
@@ -0,0 +1,271 @@
...
+    Builds a graph from detections and extracts multiple disjoint paths.
+    """
+
+    def __init__(
+        self,
+        path_overlap_penalty=40,
+        iou_weight=0.9,
+        dist_weight=0.1,
+        size_weight=0.1,
+        conf_weight=0.1,
```
soumik12345: Please add type hints.

trackers/core/ksp/solver.py
```diff
@@ -0,0 +1,271 @@
...
+    def append_frame(self, detections: sv.Detections):
+        """
+        Add detections for a new frame to the buffer.
+
+        Args:
+            detections (sv.Detections): Detections for the frame.
+        """
+        self.detection_per_frame.append(detections)
+
+    def _get_center(self, bbox):
```
soumik12345: Missing return type hint

trackers/core/ksp/solver.py
```diff
@@ -0,0 +1,271 @@
...
+            nodeV (TrackNode): Target node.
+
+        Returns:
+            float: Edge cost.
+        """
+        bboxU, bboxV = nodeU.bbox, nodeV.bbox
+        conf_u, conf_v = nodeU.confidence, nodeV.confidence
+
+        center_dist = np.linalg.norm(self._get_center(bboxU) - self._get_center(bboxV))
+        iou_penalty = 1 - self._iou(bboxU, bboxV)
```
soumik12345: Instead of using custom IoU computation for `KSPTracker`, if you can come up with a way to optimize `box_iou_batch`, not only all the other trackers will benefit, but also the community will benefit in general 🙂

trackers/core/ksp/solver.py
```diff
@@ -0,0 +1,271 @@
...
+            max(a[1], b[1]),
+            min(a[2], b[2]),
+            min(a[3], b[3]),
+        )
+        inter = max(0, x2 - x1) * max(0, y2 - y1)
+        area_a = (a[2] - a[0]) * (a[3] - a[1])
+        area_b = (b[2] - b[0]) * (b[3] - b[1])
+        return inter / (area_a + area_b - inter + 1e-6)
+
+    def _edge_cost(self, nodeU: TrackNode, nodeV: TrackNode):
```
soumik12345: Missing return type hint

trackers/core/ksp/tracker.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def reset(self) -> None:
+        """
+        Reset the KSPTracker and its solver state.
+
+        This clears all buffered detections and resets the underlying solver.
+        """
+        self._solver.reset()
+
+    def __update(self, detections: sv.Detections) -> sv.Detections:
```
soumik12345: Please make it `_update`

trackers/core/ksp/tracker.py
```diff
@@ -0,0 +1,219 @@
...
+        Args:
+            detections (sv.Detections): Detections for the current frame.
+
+        Returns:
+            sv.Detections: The same detections passed in.
+        """
+        self._solver.append_frame(detections)
+        return detections
+
+    def __assign_tracker_ids_from_paths(
```
soumik12345: Please make it `_assign_tracker_ids_from_paths`

soumik12345 (changes requested):
Hi @Ashp116, thanks for adding the documentations, they seem to be sufficient. However, I recommend that we also cite [Tracklet Association Tracker](https://arxiv.org/pdf/1808.01562) and just mention that ideas from this paper served as inspiration for the modified edge cost function. As for an in-depth explanation, would you be open to writing a blog post on [blog.roboflow.com](https://blog.roboflow.com/) on `KSPTracker`?

As for the docs UI style, I think that is out of scope for this PR. Feel free to open an issue and we can discuss further, I am sure @capjamesg would love discuss this.

Overall, I the PR is shaping up great; I left some minor feedbacks.

trackers/core/ksp/tracker.py
Ashp116: Are we not making `track` call the model to inference the frame? When we send a video source path, it is assumed we will resolve for the model detections. With this in mind and the various ways to inference; a function is more versatile.

Ashp116:
Hi @soumik12345 ,

Thank you for reviewing my PR. I’ve applied all the requested changes **except for two**:

1. **Notebook-style code**: I’d prefer to create the notebook after the full KSP tracker is finalized. This way, I can submit a dedicated PR to that repository, and once that’s merged, I’ll update this PR accordingly.

2. **sv.Detections comment**: Regarding your note that _“this should just be a list of `sv.Detections`, not a function”_—I’m not entirely sure how to resolve this. The current implementation is designed to use a function as a single entry point, which I believe helps encapsulate the logic cleanly. I'd appreciate your thoughts on how you'd suggest restructuring it.



**Major Improvement: Disjoint Paths in KSP**

I’ve made a **significant improvement** to how disjoint paths are handled in the K-Shortest Paths (KSP) algorithm. Previously, there were frequent tracker ID switches, especially when objects entered or exited the scene.

To address this, I implemented the approach described in the [paper](http://vision.stanford.edu/teaching/cs231b_spring1415/papers/Berclaz-tracking.pdf):

> *“We do this by introducing two additional nodes, υ<sub>source</sub> and υ<sub>sink</sub>, into our graph, which are linked to all the nodes representing positions through which objects can respectively enter or exit the area, such as doors or borders of the camera field of view.”*

Following this, I introduced support for **entry/exit regions**—including both custom-defined regions and automatic frame-border regions. These act like "doors" through which objects can enter or leave the scene, enabling the graph to represent disjoint paths more accurately.

This change dramatically improves tracking stability. Tracker switches are now significantly reduced, and the disjoint paths are now handled at the **graph-building stage**, while `_assign_tracker_ids_from_paths` is now solely responsible for assigning tracker IDs.

I’ll attach videos showing the improvements.

Also, I'd love to write a blog post explaining how this KSP tracker works, including the recent changes. What’s the best way to go about contributing a blog post?

Here is the [collab notebook](https://colab.research.google.com/drive/1eAJ7HzxWpWtY3NAwMtEYhJXT82lc2VZQ?usp=sharing)

Thanks again!

## Vidoes

**Example**: Look at the person labeled with tracker ID `18` in the **old KSP** and `19` in the **new KSP**.  
You can clearly see that the new KSP has a **much more stable tracker**.

### Old KSP

https://github.com/user-attachments/assets/ff91dbb2-37be-4f18-be8a-fde60ee3254d

### New KSP

https://github.com/user-attachments/assets/77113be6-8451-40f9-b860-cdf6cfcaeb3a

PR #158, opened 2026-01-08
Bytetrack adjustments

AlexBodner:
# Description

Made the corrections pointed out by Tomasz, which are:
- Default parameters now meet original bytetrack defaults
- RE-ID is now default as None.
- added additional parameters that were in original bytetrack  and modified names (hardcoded there), which are: 
 

    - [ ] Modified: minimum_iou_threshold-> minimum_iou_threshold_high_conf,  high_prob_boxes_threshold-> high_conf_boxes_threshold

    - [ ] added: minimum_iou_threshold_low_conf, minimum_iou_threshold_unconfirmed (these 2 basically now give the chance to the user of having different cost limits to each different association step), low_conf_boxes_lower_bound (the minimum value we will use to consider low confidence boxes, hardcoded to 0.1 in original)

-  Fixed how we use the minimum_iou_threshold in each association step. Now that is not only the limit, but we first modify the IoU matrix to not match those with higher lower IoU than the limit, imitating what the original does with the lap library, but building a similar solution around scipy linear_sum_assignment. This can be found in trackers/utils/bytetrack_utils.py
- Added score fusion to IoU based association, which weights IoU by object confidence. Can also be found in trackers/utils/bytetrack_utils.py
- Now in association step we internally minimize, which matches the implementation of the original bytetrack (just to avoid confussions, same behaviour)
- Renamed trackers list to tracks
- Changed behaviour with uncofirmed tracks to match the original implementation. We now put unmatched high confidence detections in an unconfirmed list, which is then matched in t+1 with the remaining high confidence detections, if there is a match we then add them to the tracks list.

Also merged main into this branch in order to fix the supervision version problem with 'box_iou_batch' import. 

## Type of change

-   [ ] New feature (non-breaking change which adds functionality)

Been testing it with an additional notebook that is not pushed. Have been viewing if results are ok and some prints to see if values were as expected. We still need to add a proper test section for trackers.

Can also be run in https://colab.research.google.com/drive/1nLNHT-MN03wuuzVOZAI934vy_-KucqPm?usp=sharing

Evaluation of results still doesnt match with ByteTrack metrics from McByte paper, but metrics in this version are overall LOWER than the ones in the first PR. 

image

image

image

## Docs

-   [ ] Docs updated?  Only docstrings of functions that have changed its parameters and main ByteTrackTracker clas


SkalskiP:
I'm closing this PR, as we already merged ByteTrack via https://github.com/roboflow/trackers/pull/174.

PR #208, opened 2026-01-29
Fix: Prevent in-place mutation of Detections input in ByteTrack

Aaryan2304:
## Summary
This PR fixes a side-effect in `ByteTrackTracker.update()` where the input `sv.Detections` object was modified in-place, overwriting existing `tracker_id` values with `-1`. This behavior caused data loss for pipelines that reuse the `detections` object after the tracking step.

## Changes
- Removed redundant in-place initialization: `detections.tracker_id = -np.ones(len(detections))` (line 111)
  - This was unnecessary because all outputs are built via `deepcopy()` calls in `_update_detections()`, the unmatched loop, and `_spawn_new_trackers()`.
- Updated the empty tracks/detections edge case (line 103-104) to return a `deepcopy()` instead of mutating and returning the input directly.

## Verification
- All ruff checks pass (linting and formatting)
- Type checking (mypy) passes with no errors
- Verified locally that the input `detections` object retains its original values after calling `update()`
- Confirmed that returned tracking results are correct and contain expected IDs

## CLA Signing
I have read the CLA Document and I sign the CLA.


Borda (approved):

SkalskiP:
Hi @Aaryan2304 👋🏻 Thanks a lot for your interes in trackers and for submitting this PR. We are planning `trackers-2.2.0` release for this week and I'll make sure we will include this fix.

Aaryan2304:
> Hi @Aaryan2304 👋🏻 Thanks a lot for your interes in trackers and for submitting this PR. We are planning `trackers-2.2.0` release for this week and I'll make sure we will include this fix.

Thanks @SkalskiP ! Appreciate you including this in the next release. 
Let me know if you need any changes to the PR before then.

trackers/core/bytetrack/tracker.py
```diff
@@ -110,17 +110,17 @@ def update(
...
         updated_detections: list[
             sv.Detections
         ] = []  # List for returning the updated detections with its new assigned track id # noqa: E501
 
         # Predict new locations for existing tracks
         for tracker in self.tracks:
             tracker.predict()
-        # Assign a default tracker_id with the correct shape
-        detections.tracker_id = -np.ones(len(detections))
+
```
SkalskiP, line 123: @Aaryan2304 why removing this line?

SkalskiP: @AlexBodner feels like we need to assign a default `tracker_id` before `_get_high_and_low_probability_detections`. Am I right?

SkalskiP:
Hi @Aaryan2304 👋🏻

It looks like the same bug is also partially present in SORT.

trackers/core/sort/tracker.py lines 133-135
```python
if len(self.trackers) == 0 and len(detections) == 0:
    detections.tracker_id = np.array([], dtype=int)
    return detections
```

trackers/core/bytetrack/tracker.py lines 112-114
```python
if len(self.tracks) == 0 and len(detections) == 0:
    detections.tracker_id = np.array([], dtype=int)
    return detections
```

trackers/core/bytetrack/tracker.py
AlexBodner: Seems like it can be safely removed, as when we copy detections and append to updated_detections we always overwrite the tracker_id. I will run a benchmark on this

SkalskiP: So this line was redundant?

AlexBodner: from what im seeing now yes

AlexBodner:
@SkalskiP ran soccernet benchmark over this branch and gives same result + from my overview seems like an adequate change

SkalskiP (approved):

SkalskiP (approved):

SkalskiP:
@Aaryan2304 I'm merging the PR. Thanks a lot for your contribution! We may need similar PR for SORT.

Aaryan2304:
> @Aaryan2304 I'm merging the PR. Thanks a lot for your contribution! We may need similar PR for SORT.

Ok, I'll make the necessary changes and raise a PR accordingly.

PR #228, opened 2026-02-05
docs: improve doctest support and code copy functionality

SkalskiP:
## Summary

- Adopt supervision's `pycon_copy.js` for cleaner code copying (strips `>>>` and `...` prompts, removes output lines)
- Enable pytest doctest support with proper configuration
- Convert docstring examples to runnable doctest format
- Add mkdocs hook to strip `# doctest: +SKIP` directives from rendered documentation


Borda (approved):
nice one with directive removal

Issue #260, opened 2026-02-16
cli: add hyperparameter tuning for trackers


SkalskiP:
## Description

Add a `trackers tune` CLI subcommand that runs hyperparameter optimization for any registered tracker. The tuner takes pre-computed detections (MOT format) and ground-truth annotations, repeatedly instantiates the tracker with different parameter configurations proposed by Optuna, evaluates each configuration using the existing `evaluate_mot_sequences` engine, and reports the best-found parameters.

```bash
# Basic usage — tune ByteTrack on MOT17 with 100 trials
trackers tune \
  --tracker bytetrack \
  --gt-dir path/to/gt/ \
  --detections-dir path/to/detections/ \
  --n-trials 100

# Pick objective metric and metric families to compute
trackers tune \
  --tracker sort \
  --gt-dir path/to/gt/ \
  --detections-dir path/to/detections/ \
  --metrics HOTA MOTA IDF1 \
  --objective HOTA \
  --n-trials 200
```

## Search space definition

Each tracker declares a `search_space` class variable alongside its `__init__` parameters.

```python
class ByteTrackTracker(BaseTracker):
    search_space: ClassVar[dict[str, dict]] = {
        "lost_track_buffer": {"type": "randint", "range": [10, 61]},
        "track_activation_threshold": {"type": "uniform", "range": [0.5, 0.9]},
        "minimum_iou_threshold": {"type": "uniform", "range": [0.05, 0.5]},
        "high_conf_det_threshold": {"type": "uniform", "range": [0.3, 0.8]},
    }
```

## Tasks

- [ ] Define `search_space: ClassVar[dict[str, dict]]` on `SORTTracker` and `ByteTrackTracker` with sensible ranges for each tunable parameter. Validate that every key in `search_space` matches a parameter name in `__init__`.
- [ ] Implement a `Tuner` class that wraps an Optuna study: converts `search_space` to Optuna distributions, runs trials calling the objective function.
- [ ] Register the `tune` subcommand in `trackers/scripts/__main__.py` with arguments: `--tracker`, `--gt-dir`, `--detections-dir`, `--metrics`, `--objective`, `--n-trials`, `--output`, `--resume`.
- [ ] Add `optuna` as an optional dependency under `trackers[tune]` in `pyproject.toml`.
- [ ] Document usage: expected input format for detections-dir and gt-dir, how to generate detections with `trackers track`, example commands, and the meaning of output files.

omkar-334:
[...]

PR #264, opened 2026-02-18
Add search spaces to trackers

omkar-334:
## What does this PR do?
Part of #260
Adds search space support to trackers

## Testing

## Checklist

- [ ] I have commented my code where necessary, particularly in hard-to-understand areas
- [ ] I have updated the documentation accordingly (if applicable)


omkar-334:
@SkalskiP may need your help on defining the search space limits... What do you think about current limits?

SkalskiP:
@tstanczyk95 can you take a look at those search ranges?

tstanczyk95:
@omkar-334 @SkalskiP my recommendations are as follows:

### ByteTrack:

- **lost_track_buffer** - currently `[10, 61]` - go until 90 inclusive. And I would say it doesn't make much sense to check each random int from that range. Go for the differences of 5 or 10.
-  **track_activation_threshold** - currently `[0.5, 0.9]` - It is ok. But mind that in the original implementation, ByteTrack does `track_activation_thresh = high_conf_det_threshold + 0.1`. See it [here](https://github.com/FoundationVision/ByteTrack/blob/main/yolox/tracker/byte_tracker.py#L154).
- **minimum_iou_threshold** - currently `[0.05, 0.5]` - it's very good, that's what I would generally expect. And I assume we talk here about IoU overlap/similarity, not IoU distance (1-IoU as in the original ByteTrack). If it doesn't cost much, we can make extra experiments and even extend this range to [0.05, 0.7]. Does it change every 0.05?
-  **high_conf_det_threshold** - currently `[0.3, 0.8]` - under normal circumstances, I would say that in general 0.3 is too low, but I assume we want to use it also on sequences outside the currently considered datasets, which might have lower quality detections? Also, ensure the search is very fine-grained here. In original ByteTrack, it was a matter of 0.01 precision, see [here](https://github.com/FoundationVision/ByteTrack/blob/main/yolox/evaluators/mot_evaluator.py#L146). 
- **minimum_consecutive_frames** - currently `[1, 5] `- in my opinion 5 is too much. Even 4 might be too much. I mean, I don't think it would help much, maybe even deteriorate the performance. I would aim mostly for 1-3. Yet if it's not costly, 4 and 5 could be considered as an supplementary option, e.g. if the user wants more extensive search, etc.

### SORT:

- **lost_track_buffer** - currently `[10, 61]` - as in the ByteTrack remarks above.
- **track_activation_threshold** - currently `[0.1, 0.6]` - I would recommend to extend this search space - even until 0.9 as for ByteTrack above.
- **minimum_consecutive_frames** - currently `[1, 5]` - as in the ByteTrack remarks above.
- **minimum_iou_threshold** - currently `[0.1, 0.6]` - I recommend starting the range with 0.05 and seeing how it behaves. Then, similar remark as with ByteTrack above: try until 0.7.

### Extra remark:
Mind that SORT was originally developed back in 2016, when the detections were of significantly lower quality than today or even than when ByteTrack was introduced in 2021. The point is, with stronger detectors these days, we can expect more from the detection quality and thus we can assume higher detection confidence than originally - especially in case of SORT.

trackers/core/bytetrack/tracker.py
```diff
@@ -45,6 +45,14 @@ class ByteTrackTracker(BaseTracker):
 
     tracker_id = "bytetrack"
 
+    search_space: ClassVar[dict[str, dict]] = {
```
SkalskiP, line 48: taking @tstanczyk95 comment: https://github.com/roboflow/trackers/pull/264#issuecomment-3944893749, I suggest updating this search space

```python
search_space: ClassVar[dict[str, dict]] = {
      "lost_track_buffer": {"type": "randint", "range": [10, 91]},
      "track_activation_threshold": {"type": "uniform", "range": [0.1, 0.9]},
      "minimum_iou_threshold": {"type": "uniform", "range": [0.05, 0.7]},
      "high_conf_det_threshold": {"type": "uniform", "range": [0.3, 0.8]},
      "minimum_consecutive_frames": {"type": "randint", "range": [1, 4]},
  }
```

trackers/core/sort/tracker.py
```diff
@@ -41,6 +43,13 @@ class SORTTracker(BaseTracker):
 
     tracker_id = "sort"
 
+    search_space: ClassVar[dict[str, dict]] = {
```
SkalskiP, line 46: taking @tstanczyk95 comment: https://github.com/roboflow/trackers/pull/264#issuecomment-3944893749, I suggest updating this search space

```python
search_space: ClassVar[dict[str, dict]] = {
    "lost_track_buffer": {"type": "randint", "range": [10, 91]},
    "track_activation_threshold": {"type": "uniform", "range": [0.1, 0.9]},
    "minimum_consecutive_frames": {"type": "randint", "range": [1, 4]},
    "minimum_iou_threshold": {"type": "uniform", "range": [0.05, 0.7]},
}
```

trackers/core/base.py
SkalskiP, line 236: Could be cool to add `search_space: ClassVar[dict[str, dict] | None] = None` here. This will probably require making updates in few other places.

SkalskiP, line 220: We could update the docstring here

```
Abstract tracker with auto-registration via tracker_id class variable.
Subclasses that define `tracker_id` are automatically registered and
become discoverable. Parameter metadata is extracted from __init__ for
CLI integration.
Attributes:
    tracker_id: Unique identifier for the tracker. Subclasses must define
        this to be registered.
    search_space: Hyperparameter search space for tuning. Each key must
        match an `__init__` parameter. Values are dicts with `type`
        (`"randint"` or `"uniform"`) and `range` (`[low, high]`).
```

test/core/test_registration.py
```diff
@@ -268,6 +269,102 @@ def test_tracker_params_have_descriptions(self, tracker_id: str) -> None:
         assert has_descriptions
 
 
+class TestSearchSpaceValidation:
+    """Tests for search_space ClassVar validation in __init_subclass__."""
+
+    def test_bytetrack_has_search_space(self) -> None:
```
SkalskiP: Feels like we could drop `test_bytetrack_has_search_space` and `test_sort_has_search_space` tests.

SkalskiP (changes requested):

SkalskiP:
Hi @omkar-334! 👋🏻 This looks really good. I left few comments. Small stuff.

trackers/core/base.py
omkar-334: yes, i missed this. thanks!

omkar-334:
> @omkar-334 @SkalskiP my recommendations are as follows:
> 
Thank you for the recommendations @tstanczyk95 !



> Hi @omkar-334! 👋🏻 This looks really good. I left few comments. Small stuff.

Thanks! I've addressed your comments.

SkalskiP (approved):

SkalskiP:
Hi @omkar-334, this looks good. Since this task will be split into several parts, I’ve created a new branch `feat/cli/tune-hyperparameter-optimization` where we’ll merge the individual pieces as they are completed. Once everything is in place, we’ll merge it into `develop`.

PR #293, opened 2026-02-24
added track-id filter to demo

AlexBodner:
## What does this PR do?

Added track-id filter to HF demo. In local couldnt render the videos when they finished processing, so please try that that works, but could download them and results were ok.

## Type of Change

## Testing

Tested same border cases as in CLI.


SkalskiP:
Hi @AlexBodner 👋🏻 ! Thanks for this. I made small changes improving the examples. Merging.

PR #310, opened 2026-03-05
Feat/core/refactor tracklet kalman filter

### LINKED ISSUE #359: Unify internal tracker container name across trackers (tracks vs trackers)
Hi, I noticed that the internal container for active track objects is not named consistently across trackers:

SORTTracker uses self.trackers
ByteTrackTracker uses self.tracks
OCSORTTracker uses self.tracks

Could this be unified into a single interface?

This inconsistency makes it harder to build generic utilities on top of the library, because downstream code has to special-case different tracker classes just to access active tracks.

It would be great to standardize on one attribute name (tracks or trackers) and possibly keep the old name as an alias for backward compatibility.

Thanks for considering it.

AlexBodner:
## What does this PR do?

Refactors KalmanBoxTracker to Tracklet classes so decouples Kalman Filter logic from the tracklet, also creating a unique Kalman Filter that will be used by all the algorithms based on the OC-SORT paper implementation of the KF. This also supports swappable kalman filter state coordinates, using the abstraction StateEstimator.

Beware that:

- OC-SORT and SORT metrics remain the same BUT ByteTrack ones has small changes (at least with default parameters) because it now uses the same kalman filter as OC-SORT (which has slight differences in implementation that make it better in OC-SORT) but here it drops 0.1%. Initialization for bytetrack remains the same as we now support custom initializations per tracker.

- ByteTrack has a behaviour difference wrt other trackers, to match original it initializes tracklets number_of_successful_consecutive_updates to 1, while in the other ones it is initalized to 0.  Also, before this PR our ByteTrack used  number_of_updates, instead of counting consecutive updates. Changing this gives different performance (+0.4% HOTA DanceTrack and -0.07% SportMOT reported in test cases and throwing the error). I think this change makes sense for the tracker, so we might have to consider redoing the parameter tuning to have results again with this change or roll this back to keep the same behaviour as the previous version of bytetrack.

## Type of Change

- New feature (non-breaking change that adds functionality BUT modifies behaviour)

---

https://github.com/user-attachments/assets/219acc15-c6c5-4bf2-93d2-8c1b5523f4f1

---

closes #359


Borda (dismissed):

docs/learn/state-estimators.md
```diff
@@ -0,0 +1,255 @@
...
+- In **SportsMOT**, for OC-SORT and ByteTrack, the StateEstimator doesn't affect the performance, while for SORT XYXYStateEstimator gives a small advantage of ~2% HOTA with default parameters and 0.4% when tuning both.
+- In **MOT17**, with default parameters XYXYStateEstimator performs slightly better than XCYCSRStateEstimator with SORT and ByteTrack with up to 0.7% better results, but for OC-SORT XCYCSRStateEstimator gives 0.2% better HOTA. When tuning parameters, XCYCSRStateEstimator performs the best with all the trackers by a small margin, ranging in 0.2-0.4% HOTA.
+
+But lets visualize where these differences are, here is an example where using XCYCSR State Estimator associates an occluded track correctly, while using XYXY changes the ID:
+
+<div style="display: flex; justify-content: center;">
+  <video style="width: 50%; height: auto;" controls>
+    <source src="../../assets/SNMOT-118-single-track-xyxy-vs-xcycsr.mp4" type="video/mp4">
```
Borda:
```suggestion
    <source src="https://github.com/user-attachments/assets/219acc15-c6c5-4bf2-93d2-8c1b5523f4f1" type="video/mp4">
```

PR #337, opened 2026-03-27
Bot-SORT refactored + Original ByteTrack alignment

AlexBodner:
## What does this PR do?

Refactor of Bot-SORT from Tomasz with kalman box tracker class to kalman filter separetely from tracklet. That doesnt change metrics.

Then aligned Bot-SORT thing inherited from "overpowered" ByteTrack to how it was done in the original one:
- The original ByteTrack performs score fusing, based on the detection confidence. It happens during the first association stage and during the unconfirmed detections matching (towards initiating new tracklets)
- Original ByteTrack stores lost_stracks. We had similar mechanisms in ByteTrack in Trackers with maximum_frames_without_update, but the original one also makes additional distinction. It only considers lost_stracks in the first association step and not in the second association step. 
- Now we give 1 frame to confirm unconfirmed tracks (in step 3)

---

closes #334


Borda:
@AlexBodner, could you pls try to rebase or cherry-pick the commits? It seems that merging has cumulated too many unrelated commits 🦝

PR #345, opened 2026-04-01
IoU variants

AlexBodner:
## What does this PR do?

Added elegible IoU variants to trackers!
Now you can call any tracker like tracker(iou = IoU()) or tracker(iou = GIoU()), etc.
Available variants:
- GIoU
- DIoU
- CIoU
- BIoU

## Type of Change

- Documentation update

## Testing

- [ ] I have added/updated tests for this change

**Test details:**
We compare with torchvision results and math properties that the IoUs should follow in order to check expected results.


docs/trackers/botsort.md
```diff
@@ -0,0 +1,151 @@
+---
+comments: true
```
Borda, line 2: Check the other pages, each shall also have also describtion...

docs/trackers/botsort.md
```diff
@@ -0,0 +1,151 @@
...
+        if cv2.waitKey(1) & 0xFF == ord("q"):
+            break
+
+    video_capture.release()
+    cv2.destroyAllWindows()
+    ```
+
+=== "Webcam"
+
+    ```python
```
Borda, line 71: Do we have tests for these examples?

test/utils/test_iou.py
```diff
@@ -0,0 +1,547 @@
...
+_biou = BIoU()
+_giou = GIoU()
+_diou = DIoU()
+_ciou = CIoU()
+
+
+class TestGIoUAgainstTorchvision:
+    """Compare our GIoU against torchvision.ops.generalized_box_iou."""
+
+    def test_identical_boxes(self) -> None:
```
Borda, line 48: since the only digg is input and exp. value we can use parameterize and fraction of the test name as id `pytest.param(..., id="")`

test/utils/test_iou.py
```diff
@@ -0,0 +1,547 @@
...
+        biou = BIoU(buffer_ratio=0.1).compute(boxes_1, boxes_2)[0, 0]
+        assert iou == 0.0
+        assert biou > 0.0
+
+    def test_invalid_negative_buffer_ratio(self) -> None:
+        with pytest.raises(ValueError, match="buffer_ratio must be non-negative"):
+            BIoU(buffer_ratio=-0.01)
+
+
+class TestDIoUAgainstTorchvision:
```
Borda, line 179: on the other hand, we could also parameterize the entire class since the testing logic seems identical, compared to the baseline implementation, we can have a variant as a parameter and ...
```py
  "ours", "ref_or_baseline",
  (_giou, _torchvision_giou),
  (_ciou, _torchvision_ciou),
...
```

test/utils/test_iou.py
```diff
@@ -0,0 +1,547 @@
...
+
+
+_iou = IoU()
+_biou = BIoU()
+_giou = GIoU()
+_diou = DIoU()
+_ciou = CIoU()
+
+
+class TestGIoUAgainstTorchvision:
```
Borda, line 45: is this duplicated?

trackers/core/botsort/tracklet.py
```diff
@@ -0,0 +1,230 @@
...
+import numpy as np
+
+from trackers.utils.base_tracklet import BaseTracklet
+from trackers.utils.state_representations import (
+    BaseStateEstimator,
+    XCYCWHStateEstimator,
+)
+
+
+class BoTSORTTracklet(BaseTracklet):
```
Borda, line 20: lets ahve adding tracker in separate PR

Borda (changes requested):

AlexBodner:
Moving this to PR: https://github.com/roboflow/trackers/pull/403 @Borda

docs/trackers/botsort.md
AlexBodner: no, we neved had tests for examples and would be nice to have. How could we do so that the test gets the code from the mds? is that a thing?

test/utils/test_iou.py
AlexBodner: nice suggestion, addressed in commit:4e293dd8c2cf8f1beb4605488100487c6bf81cd9

test/utils/test_iou.py
AlexBodner: also addressed in commit:4e293dd8c2cf8f1beb4605488100487c6bf81cd9

PR #373, opened 2026-04-23
Return all alive tracks

### LINKED ISSUE #105: [feature request] Returning tracks that are alive but occluded
I've been working on integrating your SORT/ Deep SORT trackers into an application, but I've found a confusing design decision. It looks you expect us to use the output of `tracker.update(detections)` to represent the tracks present on the current frame. `update` only returns the subset of tracks that match to the detections that were available for the given frame. This "hides" tracks that are still alive but may be temporarily occluded or spuriously not detected. 

**Does it make sense to provide a method that returns all alive tracks given the current tracker state?**

omkar-334:
## What does this PR do?

Adds a `tracked_objects` property on `BaseTracker` (and implementations in the individual tracker) that exposes every alive track with its Kalman-predicted bounding box, including tracks that were not matched to a detection on the current frame (occluded or detector-missed).                                                                   
                                                                                                                                               
  Matches the API shape proposed by @SkalskiP in the issue thread:                                                                             
                                                                                                                                               
  ```python                                                                                                                                    
  @property                                                                                                                                    
  @abstractmethod                                                                                                                            
  def tracked_objects(self) -> sv.Detections: ...
 ```  
 
update() is untouched, so existing users see no behavior change.   
Resolves #105

## Type of Change

## Testing

<!-- Describe how you tested your changes -->

**Test details:**
<!-- Describe what tests were added/updated and what they cover -->

## Checklist

## Additional Context
<img width="1038" height="228" alt="Screenshot 2026-04-24 at 12 25 46 AM" src="https://github.com/user-attachments/assets/87d6e5b0-44e3-45e5-b858-142686028b9f" />  

all current tests pass


omkar-334:
2 thoughts about this -   
1. Will the implementation of this function change per tracker? If not, we can just implement it in the `BaseTracker` and use 1-line hooks in the actual trackers.... Currently the implementations are the same.  

Base tracker - 
```python
@property
def tracked_objects(self) -> sv.Detections:
    alive = self._alive_tracklets()
    if not alive:
        empty = sv.Detections.empty()
        empty.tracker_id = np.array([], dtype=int)
        return empty
    xyxy = np.array([t.get_state_bbox() for t in alive], dtype=np.float32)
    ids = np.array([t.tracker_id for t in alive], dtype=int)
    return sv.Detections(xyxy=xyxy, tracker_id=ids)

@abstractmethod
def _alive_tracklets(self) -> list: ...
```  

trackers - 
```python
def _alive_tracklets(self): return self.tracks   # or self.trackers for SORT
```

omkar-334:
2. with this PR, the user can currently filter for mature tracks like this - 
```python
alive = tracker.tracked_objects
mature = alive[alive.tracker_id != -1]
  ```  
  Do you think this is fine, or do we add a parameter to do this in the function itself

File: trackers/core/bytetrack/tracker.py [hunk 1]
```diff
@@ -259,3 +259,7 @@ def reset(self) -> None:
         """
         self.tracks = []
         ByteTrackKalmanBoxTracker.count_id = 0
+
+    def _alive_tracklets(self) -> list:
+        """Return the tracks currently considered alive."""
+        return self.tracks
```
Borda: why is this private method needed since it is always tacked and not even a copy?

omkar-334: @Borda The private function to unify the naming convention - it's there only to bridge `self.trackers` in SORTTracker vs `self.tracks` in ByteTrackTracker/OCSORTTracker.   
Issue - #359

omkar-334: if the issue is resolved, we dont need it. can just use `self.tracks`

Borda: @omkar-334 the refactoring PR #310 was merged

omkar-334: @Borda thanks! rebased the PR, now we dont need this private method anymore.  
However, the tests are failing because of the bytetrack refactor. I added details in the #376 pr

omkar-334:
@Borda i've updated this PR now that #376 is merged. can you review this?

Borda (approved):

PR #415, opened 2026-05-13
Fix ByteTrack handling of detections without confidence

cdeil:
## What does this PR do?

Make `ByteTrackTracker.update()` treat `detections.confidence is None` as all-ones confidence, matching SORT, OC-SORT, and BoT-SORT behavior.

Previously, ByteTrack defaulted missing confidence to zeros, which routed every detection into the low-confidence bucket. Since ByteTrack only starts new tracks from unmatched high-confidence detections, valid `sv.Detections(xyxy=...)` inputs could silently return only `tracker_id=-1` forever.

## Type of Change

I think this could be considered a bug fix or at least consistency improvement.

## Testing

## Checklist

## Note

As mentioned previously I want to do a Dart clone of Trackers so that I can use it in my mobile app offline.
To avoid rework I asked a coding agent to review the Python package with Claude Opus 4.7 and GPT 5.5 and it suggested several fixes and improvements.

See here: https://gist.github.com/cdeil/0807f9b4f6e668f4277614822bc8b52f

@Borda - maybe you could skim the list and let me know if you think which/any are actual issues / worth doing. And if you want me to send PRs like this one or just prompt yourself. I'm not very familiar with this, so unfortunately probably won't be able to add much value over what the agents suggest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - **OC-SORT Observation-Centric Recovery** now uses standard `IoU` per the paper, independent of the configured `iou=` variant ([#403](https://github.com/roboflow/trackers/pull/403)).
 - **Eager division warnings on zero-area boxes** — IoU helper switched from `np.where` (eager) to `np.divide(..., where=...)` (lazy), suppressing `RuntimeWarning` under strict NumPy error settings ([#403](https://github.com/roboflow/trackers/pull/403)).
 - **CLI argparse crash on `BaseIoU` parameter** — `iou=` is now excluded from argparse auto-discovery; the variant must be set programmatically ([#403](https://github.com/roboflow/trackers/pull/403)).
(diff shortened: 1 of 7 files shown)


Borda:
> As mentioned previously I want to do a Dart clone of Trackers so that I can use it in my mobile app offline. To avoid rework I asked a coding agent to review the Python package with Claude Opus 4.7 and GPT 5.5 and it suggested several fixes and improvements.
> 
> See here: [gist.github.com/cdeil/0807f9b4f6e668f4277614822bc8b52f](https://gist.github.com/cdeil/0807f9b4f6e668f4277614822bc8b52f)

With a quick look, it looks reasonable, but I have not checked the details yet...
TO move forward, I would suggest copying the first as an issue so we can better reference and track progress over your findings 🦝

Borda (dismissed):

AlexBodner:
@Borda we should apply the change only to bytetrack/tracker.py (and propagate to other trackers + unit test that), but the rest of the changes i dont think apply to the repo, as it is duplicating the code to dart.

cdeil:
I was asking a coding agent to develop https://github.com/cdeil/dart_trackers in a sub-folder and it accidentally pushed those changes here. I'm sorry for that it's a git screw-up. I will fix it now.

cdeil:
OK there's now a single commit ad78beaf4c2bbea54d8410c8039b43d9444043f9 hope it's OK.

Borda:
Yeah, about the rest in the gist, I was about to ipen Issue only to the related findings
However lets keep this PR focus only on this chnage 🦝

Borda (dismissed):

AlexBodner:
I see that we had already applied this criteria for the other trackers, seems ready to merge

PR #419, opened 2026-05-15
Replace global class-variable track ID counters with per-instance allocators

cdeil:
## What does this PR do?

Fix duplicate track IDs when multiple tracker instances of the same type are used in one process.

Track IDs are the labels returned in `detections.tracker_id`, for example person A gets `tracker_id=0`, person B gets `tracker_id=1`, and the next new track should get `tracker_id=2`. Previously, that "next ID" counter lived on the internal tracklet class, so all `SORTTracker()` instances shared it.

That meant resetting one tracker could accidentally affect another live tracker:

```python
tracker_a = SORTTracker()
tracker_b = SORTTracker()

# tracker_a creates a live track with ID 0
tracker_a.update(detections_for_person_a)

# tracker_b resets the shared SORTTracklet counter back to 0
tracker_b.reset()

# tracker_a later creates another live track
# before this PR, that second live track could also get ID 0
tracker_a.update(detections_for_person_a_and_b)
```

This PR moves the counter to each tracker instance. `tracker_b.reset()` still restarts IDs from zero for `tracker_b`, but it no longer changes the IDs that `tracker_a` will allocate.

This can affect normal scripts and notebooks, not only multi-threaded applications. It is enough to have two trackers of the same type alive at once, for example in multi-camera tracking, class-specific tracking, side-by-side config comparisons, services handling multiple requests, or notebook experiments.

## Type of Change

I think this is a bug fix with a small internal breaking change: `BaseTracklet.count_id` / `Tracklet.get_next_tracker_id()` were removed because tracklets no longer own ID allocation.

## Testing

Red/green TDD:

- Added `test_tracker_instances_do_not_share_id_allocators`, which failed for SORT, ByteTrack, OC-SORT, and BoT-SORT with duplicate live IDs `[0, 0]`.
- Applied the per-instance allocator fix and reran the regression plus existing reset/tracklet contracts successfully.
- Ran the full pytest suite and ruff checks successfully.

## Checklist


AlexBodner:
@cdeil thanks for this PR!. @Borda this is something i wanted to do some time ago, so its great. the only problem is that it will generate conflicts with CBIou and mcbyte. So I would first merge those PRs, then extend this one to cover those trackers and finally merging this PR

Borda (approved):

PR #420, opened 2026-05-15
Unify lost track buffer semantics

cdeil:
## What does this PR do?

Unify `lost_track_buffer` semantics across SORT, ByteTrack, OC-SORT, and BoT-SORT, and prevent positive buffers from collapsing to zero frames at low frame rates.

Previously every tracker scaled the user-facing buffer with:

```python
int(frame_rate / 30.0 * lost_track_buffer)
```

That floors fractional results. For example, `frame_rate=10` and `lost_track_buffer=1` produced `maximum_frames_without_update == 0`, so a confirmed track could be deleted on the first missed detection even though the caller requested a one-frame buffer.

There was also a tracker inconsistency at the exact expiry boundary. SORT, ByteTrack, and BoT-SORT removed tracks when `time_since_update == maximum_frames_without_update`, while OC-SORT kept them until the value was exceeded. This PR makes the public meaning consistent: `lost_track_buffer=N` keeps a confirmed track alive for exactly `N` scaled missed frames and removes it on the next miss.

This matters for low-FPS and mobile streams, where small but valid buffers are common and premature deletion can cause avoidable ID switches after a single detector miss.

## Type of Change

This is a bug fix with a small breaking change for invalid configuration: all trackers now raise `ValueError` when `lost_track_buffer < 1` or `frame_rate <= 0` / non-finite. Those values previously initialized but produced nonsensical lifecycle behavior, including immediate pruning.

Not sure if this is the semantics you want. E.g. requiring `lost_track_buffer > 0` could also be reasonable. And for `maximum_frames_without_update` consistently putting `<` instead of `<=` would also be reasonable. But overall I'm +1 to make the trackers fully consistent here since it's already difficult to understand precisely what the parameters mean for casual/new users. Maybe a small docs addition/illustration would help.

## Testing

Red/green TDD:

- Added `test_track_survives_exact_lost_buffer_boundary`, which failed for SORT, ByteTrack, and BoT-SORT because they pruned at the exact buffer boundary instead of after it.
- Added `test_low_frame_rate_lost_buffer_rounds_up_to_one_frame`, which failed for all four trackers because `frame_rate=10, lost_track_buffer=1` scaled to `0`.
- Added `test_lost_buffer_configuration_rejects_non_positive_values`, which failed for all four trackers because invalid buffer settings were accepted.
- Applied the shared scaling/validation fix and reran the new regression tests, full pytest suite, and Ruff checks successfully.
- Updated the affected integration metric baselines where the corrected boundary semantics keep tracks alive for one additional missed frame.

## Checklist


AlexBodner:
@cdeil thank you for your PR!
This is something we should consider for future version, as consistency along trackers is one of the most important things, but each tracker also does its own things and handles this details differently in the original implementations, which leads to differences in metrics. That's why we opted for having the < and <= as of now. Though if we do not degrade performance and with other hyperparameters we can achieve the same results, while unifying interface interpretation this is a change that we will apply

Borda (approved):

PR #495, opened 2026-07-09
fix(utils): harden zip extraction

### LINKED ISSUE #477: Security: _extract_zip does not validate zip member paths (Zip Slip)
### Summary

`_extract_zip` in `src/trackers/utils/downloader.py` calls  `zip_file.extractall(output_dir)` without validating member paths.

A ZIP containing path traversals (`../`) or absolute paths could  write files outside the intended output directory.

### Proposed fix

Iterate `zip_file.namelist()` before extracting and reject any  member whose resolved path falls outside `output_dir`.

I have a fix ready — happy to open a PR if this is worth addressing.

### Note

I understand `trackers download` pulls from Roboflow's own GCP  bucket, so the real-world risk is low. Flagging it in case you  want to harden the utility for future use cases.

Borda:
This pull request adds security checks to ZIP extraction to prevent directory traversal attacks and introduces comprehensive tests to verify the new behavior. The most important changes are:

**Security improvements to ZIP extraction:**

* Added a new function `_validate_zip_member_path` in `downloader.py` to ensure that ZIP archive members cannot extract files outside the intended output directory, raising a `ValueError` if an unsafe path is detected.
* Updated `_extract_zip` to validate all member paths before extraction and to document the new error behavior in its docstring.

**Testing for ZIP extraction safety:**

* Added `test_extract_zip_extracts_safe_members` to verify that safe ZIP members are extracted correctly into the output directory.
* Added `test_extract_zip_rejects_unsafe_members` (with parameterization) to assert that unsafe ZIP members (such as those with `..` or absolute paths) are rejected and that no files are extracted in these cases.
* Introduced a helper function `_write_zip` for creating test ZIP archives with arbitrary members.

---

resolves #477

### DIFF
--- a/src/trackers/utils/downloader.py
+++ b/src/trackers/utils/downloader.py
@@ -11,6 +11,7 @@ import os
 import shutil
 import stat
 import zipfile
+from contextlib import suppress
 from pathlib import Path, PurePosixPath, PureWindowsPath
 
 import requests
@@ -117,10 +118,8 @@ def _safe_zip_member_parts(member_name: str) -> tuple[str, ...]:
 
 def _open_child_directory(parent_fd: int, directory_name: str) -> int:
     """Open a child directory below `parent_fd`, creating it if needed."""
-    try:
+    with suppress(FileExistsError):
         os.mkdir(directory_name, mode=0o777, dir_fd=parent_fd)
-    except FileExistsError:
-        pass
 
     directory_fd = os.open(
         directory_name,
--- a/src/trackers/utils/downloader.py
+++ b/src/trackers/utils/downloader.py
@@ -87,7 +87,28 @@ def _download_file(
     return True
 
 
+def _validate_zip_member_path(output_dir: Path, member_name: str) -> None:
+    """Reject ZIP members that would extract outside `output_dir`."""
+    if not member_name:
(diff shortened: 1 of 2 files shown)


Borda:
Measured on this machine:

- Fixed per-archive overhead
    - 0 files, 0 depth, 0-byte payload
    - plain extractall: 0.03 ms
    - safe extractor: 0.07 ms
    - extra cost: ~0.03 ms per archive

- Per-file overhead
    - 1 KiB files, depth 1
    - observed safe-extractor slope: ~0.090 ms/file
    - overhead at 100 files: 1.59 ms vs plain extractall

- Per-directory-depth overhead
    - 20 files, 1 KiB each
    - observed safe-extractor slope: ~0.282 ms/depth-level
    - overhead at depth 8: 2.05 ms vs plain extractall

- Size sensitivity
    - 20 files, depth 1
    - 1 KiB: overhead ~0.36 ms
    - 64 KiB: overhead ~1.20 ms
    - 512 KiB: overhead ~7.97 ms

Takeaway:

- The safety cost is not constant.
- It scales with:
    - number of files
    - directory depth
    - payload size as well, in this implementation

- The validation itself is cheap; most of the overhead comes from the fd-anchored extraction path and Python-level copying.

PR #504, opened 2026-07-18
fix(cbiou): keep instant-activated tracks alive on a miss

JESUSROYETH:
`CBIoUTracker.update()` reimplements BoT-SORT's lifecycle split, but it was written before #478 added the sticky-maturity guard to BoT-SORT, so the fix never reached C-BIoU. The split still decides maturity only by update count:

```python
elif track.number_of_successful_updates >= self.minimum_consecutive_frames:
    confirmed_tracks.append(track)
else:
    unconfirmed_tracks.append(track)
```

With the default `instant_first_frame_activation=True` and `minimum_consecutive_frames=2`, a first-frame track holds a real `tracker_id` but only one successful update. If the detector misses the object for a single frame (motion blur, a short occlusion), the track falls into the unconfirmed bucket and gets deleted, so the object comes back with a new ID a couple of frames later. BoT-SORT keeps the ID in this exact scenario since #478, C-BIoU still loses it.

**Changes**

- `src/trackers/core/cbiou/tracker.py`: mirror the `track.tracker_id != -1 or ...` guard (and its comment) from `botsort/tracker.py`.
- `tests/core/test_cbiou_tracker.py`: mirror all four `TestBoTSORTTrackerLifecycle` sticky-maturity regressions from #478 — survives a single miss, pruned when instant activation is off, survives two consecutive misses (confirmed → lost_tracks transition), and survives a miss across `minimum_consecutive_frames` in {1, 2, 3}.

The shared `get_alive_tracklets()` in `botsort/utils.py` already got the guard in #478, so this line in `CBIoUTracker.update()`'s own confirmed/unconfirmed split — a separate reimplementation of the same lifecycle logic — was the only missing piece. `CBIoUTracker` is the only subclass of `BoTSORTTracker`. SORT, ByteTrack and OC-SORT don't have `instant_first_frame_activation` at all, so they can't hit this scenario.

**Validation**

- The new regression tests fail on `develop` (except `mcf-1`, where a single update already reaches maturity without the guard) and pass with the fix.
- `pytest tests/core` passes (444 tests), including the C-BIoU vs BoT-SORT zero-buffer equivalence test.
- Integration suite (`-m integration -k cbiou`, DanceTrack + SportsMOT) passes with the golden metrics unchanged .. the GT-derived detections there have no missed frames, so the guard does not alter those trajectories. The behaviour change is pinned by the unit test.

### DIFF
--- a/src/trackers/core/cbiou/tracker.py
+++ b/src/trackers/core/cbiou/tracker.py
@@ -231,7 +231,12 @@ class CBIoUTracker(BoTSORTTracker):
         for track in self.tracks:
             if track.time_since_update > 1:
                 lost_tracks.append(track)
-            elif track.number_of_successful_updates >= self.minimum_consecutive_frames:
+            elif track.tracker_id != -1 or track.number_of_successful_updates >= self.minimum_consecutive_frames:
+                # Maturity is sticky: a track that already holds a real
+                # tracker_id (e.g. an instant-activated first-frame track) stays
+                # confirmed even before it reaches minimum_consecutive_frames.
+                # On a miss it is kept as a confirmed (then eventually lost)
+                # track rather than discarded as an unconfirmed one.
                 confirmed_tracks.append(track)
             else:
                 unconfirmed_tracks.append(track)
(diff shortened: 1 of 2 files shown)


JESUSROYETH:
I have read the CLA Document and I sign the CLA.

Borda (approved):

PR #526, opened 2026-07-30
perf(mcbyte): default-off precision/compile knobs

Borda:
Generalize _autocast_context(use_amp) to (use_amp, device) in the Cutie and SAM wrappers so the autocast backend follows device.type instead of a hardcoded "cuda"; gating stays cuda-only, so the default path is unchanged.

Add _apply_backend_perf_options to CutieMaskPropagator: an opt-in channels_last memory format and an opt-in torch.compile of the shape-stable encoder methods encode_image and transform_key. Cutie's InferenceCore calls network submethods directly rather than forward(), so a module-level compile is inert; segment and encode_mask stay eager to avoid recompilation on their varying object-count dimension.

Thread cutie_channels_last and cutie_compile through McByteMaskConfig. All new knobs default off, leaving the default pipeline bit-identical. Add unit tests for the autocast context, the perf-option helper, and config threading.

### DIFF
--- a/tests/core/test_mcbyte_cutie_mask_propagator.py
+++ b/tests/core/test_mcbyte_cutie_mask_propagator.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+from contextlib import nullcontext
 from typing import Any
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


tests/core/test_mcbyte_cutie_mask_propagator.py
```diff
@@ -1155,3 +1159,131 @@ def output_prob_to_mask(self, prob: Any) -> Any:
...
+    assert isinstance(context, nullcontext)
+    assert autocast_calls == []
+
+
+@pytest.mark.parametrize(
+    "device_type",
+    [
+        pytest.param("cuda", id="cuda"),
+        pytest.param("cpu", id="cpu"),
+    ],
```
Borda:
```suggestion
    ["cuda","cpu"]
```

PR #543, opened 2026-08-08
feat(cli): add trackers inspect command group

Borda:
- add `trackers inspect` exposing sam, cutie, mask-manager, and mcbyte components through INSPECT_COMPONENTS, moved from the untracked-by-docs visual_tests/ scripts into src/trackers/cli/inspect/
- rename `trackers mcbyte` to `trackers benchmark mcbyte` so every top-level key in _COMMANDS is a verb and a tracker name is an argument to one; the subcommand never shipped in a release, so this is a rename rather than a deprecation
- merge the manual and ground-truth mask-manager visualizers into one command behind a required `--mode {manual,gt}`, with _raise_for_mode_option_conflict rejecting the other mode's options by name before any SAM or Cutie import
- move the tracker-agnostic mask stack from core/mcbyte/masks/ to core/masks/ and mask_manager.py to core/masks/manager.py; nothing in it referenced McByteTracker, and its tests already drove it with DummyBoxMaskGenerator
- keep mask_association.py in core/mcbyte/, since condition_similarity_with_masks is McByte's own association method
- extract IMAGE_EXTENSIONS, parse_xyxy_box, list_selected_frame_paths, load_rgb_image, save_rgb_image, timestamped_run_dir, validate_device, and print_device_info into cli/inspect/_common.py; the remaining same-named helpers differ between components and stay local
- register the mcbyte comparison dataclass groups from _CLIParser.add_function_arguments, since one parser_class now serves every nested subcommand
- add benchmark and inspect entries to _SUBCOMMANDS and _LEGACY_ARGUMENTS, both of which are subscripted per subcommand
- default outputs to outputs/inspect/<component>/ via INSPECT_OUTPUT_ROOT instead of writing into the source tree
- defer torch, SAM, and Cutie imports into command bodies behind require_torch, which raises an install hint naming the mask extra
- add tests/cli/test_inspect.py covering registration, hyphen and underscore spellings, mode conflicts, and the shared helpers
- add docs/learn/inspect.md and its nav entry


src/trackers/cli/inspect/_common.py
```diff
@@ -0,0 +1,344 @@
...
+from trackers.core.masks.base import MaskOutput, TrackletSnapshot
+
+# Device validation is not CLI-specific, so it lives with the other device
+# helpers. Re-exported here because every inspect component reaches for it.
+
+if TYPE_CHECKING:
+    import torch
+
+IMAGE_EXTENSIONS = frozenset({".jpg", ".jpeg", ".png", ".bmp", ".webp"})
+"""Frame file suffixes recognised when listing an image directory."""
```
Borda: @copilot replace all multiline strings inside the plain module body with # comments, but do not convert the initial module dostrings

PR #559, opened 2026-08-12
fix(utils): guard Kalman gain against singular innovation covariance

AlexBodner:
**Fable false positive — closing, not merging.** The problem this guards against cannot happen, and the remaining change buys nothing measurable.

- The premise was that `np.linalg.inv` on the innovation covariance could raise `LinAlgError` and kill a track. It can't: every tracker sets `R` as a positive scaled identity (`kf.measurement_noise * 0.1`, `np.eye(4) * 0.1`, `[2:, 2:] *= 10.0`), so `S = H P Hᵀ + R ⪰ R ≻ 0` stays invertible whatever `P` does — verified with `P` rank-1, `P = 0`, and `P` at 1e-300, where `S` bottoms out at `R` and stays well-conditioned
- The test in this PR had to hand-assign `measurement_noise = np.zeros((1, 1))` to reach the singular branch, i.e. a state no tracker can produce
- It also misses the degradation that is real. `xcycsr_to_xyxy` decodes a non-positive scale as NaN (see `clamp_velocity`), and `solve` does not raise on non-finite input — `S = NaN` returns `[nan nan]` and `S = inf` returns `[0, 0]`, so the guard never fires on the path that actually breaks. Non-finite state is already caught downstream by `BaseIoU.compute()`
- That leaves the `inv` → `solve` swap on its own merits, which do not survive measurement. Accuracy improves but only at machine epsilon — median residual `|K@S − PHT|` 2.67e-15 → 1.78e-15, worst case 4.97e-14 → 5.33e-15 over 2000 realistic cases — roughly 13 orders of magnitude below the sub-pixel precision box tracking cares about. Speed is a wash: 2.680 µs vs 2.577 µs per call at the real 4×4 / 7×4 sizes (best-of-15; single-run timings at this size are noise-dominated and flip direction between trials)
- Not worth editing the Kalman hot path that #296 deliberately optimized for a change nothing downstream can observe

Worth keeping from the investigation: `S` is structurally non-singular given a positive-definite `R`, so no guard is needed there; NaN, not singularity, is the real failure mode on this path.

### DIFF
--- a/src/trackers/utils/kalman_filter.py
+++ b/src/trackers/utils/kalman_filter.py
@@ -100,9 +100,9 @@ class KalmanFilter:
 
         If z is None, the state is not updated (prediction only).
 
-        If the innovation covariance is singular — zero measurement noise, or a state covariance that has collapsed
-        after repeated updates — the gain falls back to a pseudo-inverse least-squares solution so the track survives
-        instead of raising.
+        If the innovation covariance is singular — zero measurement noise together with a state covariance that has
+        collapsed after repeated updates — the gain falls back to a pseudo-inverse least-squares solution so the track
+        survives instead of raising.
 
         Args:
             z: Measurement vector (dim_z, 1) or None for no observation.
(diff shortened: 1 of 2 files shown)


AlexBodner:
Closing per the analysis in the description: the singular-innovation case is unreachable with any tracker in the repo, and the inv → solve swap is a wash on speed and machine-epsilon on accuracy. Labelled wontfix.

PR #511, opened 2026-07-21
feat(reid): appearance ReID association via standalone reid package

AlexBodner:
## Summary

Adds optional appearance-ReID to BoT-SORT, sourcing the model stack from the standalone [`reid`](https://github.com/roboflow/re-ID) package. Trackers ships only NumPy association glue (`ReIDEncoder`, `FeatureBank`, `appearance_similarity`, `extract_detection_embeddings`) plus BoT-SORT appearance-IoU fusion. Encoder weights, preprocessing, catalog, and gallery evaluation live in `reid` and are never vendored.

This branch is merged with current `develop` (including McByte) and is independent of the earlier stacked ReID PRs (#480/#501/#502).

## Added things

- **Deps:** `trackers[reid]` optional extra → `reid>=0.1.0.dev0,<0.2` (PyPI). `reid.*` in mypy overrides; `uv.lock` updated.
- **Association glue** (`src/trackers/core/reid/`, torch-free): `ReIDEncoder`, `FeatureBank` (BoT-SORT EMA L2 before/after blend), `appearance_similarity`, `extract_detection_embeddings`.
- **Lazy import:** CLI loads `reid.ReIDModel` at use time (same pattern as detection); importing BoT-SORT stays free of `reid` / torch (isolation test).
- **BoT-SORT wiring:** appearance-IoU fusion, per-track feature bank, params `reid_model`, `reid_ema_alpha`, `reid_appearance_threshold`, `reid_proximity_threshold`.
- **CLI:** `--tracker.reid.{enable,model,device,architecture}` on `trackers track` (BoT-SORT only).
- **Tuning:** `trackers tune` and `Tuner` search `reid_appearance_threshold` and `reid_proximity_threshold` when an encoder is passed (`--reid.model`, or `reid_model` in `fixed_params`). Search-space entries gain an optional `requires` key, a `search_space` override is accepted per run, and each detection is embedded once per study instead of once per trial.
- **Tests:** association / fusion / CLI unit tests (no `--extra reid` required), `@integration` smoke with a real `reid.ReIDModel`, isolation test.
- **Docs / notebook:** `docs/learn/reid.md` (threshold guide + MOT17/SoccerNet association-local plots), `docs/api/reid.md`, install notes, BoT-SORT Appearance ReID section; `notebooks/eval_trackers_reid.ipynb` installs from PyPI.
- **CI:** unit (`ci-tests.yml`) syncs without `--extra reid`; integration (`ci-integrations.yml`) keeps `--extra reid`.

[...]

PR #571, opened 2026-08-19
feat(reid): add Deep OC-SORT adaptive appearance fusion to BoT-SORT

AlexBodner:
Stacked on #511. Adds `reid_fusion="adaptive"` to `BoTSORTTracker`: the Adaptive Weighting rule from Deep OC-SORT (arXiv:2302.11813, eqs. 4–6), selectable alongside BoT-SORT's own `min(d_iou, d_app)` fusion. Default stays `"botsort"`; behaviour is unchanged unless opted into.

https://github.com/user-attachments/assets/9e5a02ea-51b5-4421-8365-25ab09ec73f4

## What

- `fuse_adaptive_reid_association` in `trackers/core/reid/fusion.py`: `IoU + (a_w + w_b) · A_c`, where `w_b` is the capped gap between the best and second-best appearance match. Implements the difference form the authors ship as `compute_aw_new_metric` (the variant behind their published results), not the ratio form other ports carry. Only the weighting is ported — not Dynamic Appearance, the velocity term, or OCR/OOS — so it is usable by any tracker and named for what it does.
- Two new `BoTSORTTracker` parameters, `reid_appearance_weight` (`a_w`, default 0.75) and `reid_adaptive_weight_cap` (`epsilon`, default 0.5), the paper's MOT17/MOT20 values. `reid_appearance_threshold` is ignored under adaptive fusion and the docstring says so.
- Two documented departures from the reference: negative cosine similarities are clamped to zero, and appearance is gated by `reid_proximity_threshold` (the reference drops implausible pairs after matching; this integration thresholds the fused similarity directly). The discriminativeness gap is measured over all candidates *before* that gate, so a geometrically excluded candidate cannot read as similarity 0 and earn the bonus for being spatially alone.
- `docs/guides/reid.md`: fusion choice, a measured comparison, and a new section on `reid_proximity_threshold`.

## Measured

Three arms per dataset sharing detections, encoder, CMC and geometry; only appearance handling differs. Parameter columns give the full ReID settings each cell ran with, defaults included. SoccerNet and MOT17 rows use the published tuned BoT-SORT geometry from `docs/evaluations/results.md` (SoccerNet's reproduces the published 85.0 exactly); DanceTrack's tuned set equals the defaults. HOTA, best configuration per arm:

| Dataset | no ReID | `botsort` | `adaptive` | `botsort` parameters | `adaptive` parameters |
| --- | --: | --: | --: | --- | --- |
| SoccerNet test (49, oracle dets, in-domain OSNet) | 85.00 | **87.29** ¹ | 86.23 | `reid_appearance_threshold=0.075`, `reid_proximity_threshold=1.0` | `reid_appearance_weight=0.75`, `reid_adaptive_weight_cap=0.5`, `reid_proximity_threshold=1.0` |
| DanceTrack val (25, YOLOX, MSMT17 OSNet) | 53.89 | 56.11 | **57.39** | `reid_appearance_threshold=0.25`, `reid_proximity_threshold=0.5` | `reid_appearance_weight=2.4`, `reid_adaptive_weight_cap=0`, `reid_proximity_threshold=0.5` |
| MOT17 val-half (7, YOLOX, in-domain OSNet) | **69.05** | 69.00 | 68.72 | `reid_appearance_threshold=0.25`, `reid_proximity_threshold=0.5` | `reid_appearance_weight=0.75`, `reid_adaptive_weight_cap=0.5`, `reid_proximity_threshold=0.5` |

¹ `botsort` at `reid_appearance_threshold=0.075`, `reid_proximity_threshold=1.0`. It has the higher HOTA and IDF1 (83.13 vs 82.52) but 4 564 ID switches against adaptive's 1 692 and geometry's 2 523: `min(d_iou, d_app)` takes the cheapest candidate each frame and flips between same-kit players (two thirds of its identity changes revert within ten frames on SNMOT-120). Adaptive commits and holds. Neither dominates; adaptive is the rule for stable ids, which is the usual reason to enable appearance.

Adaptive cuts ID switches on SoccerNet (2523 → 1692, −33%) and DanceTrack (−12%); `botsort` fusion raises them on all three. Most of the SoccerNet HOTA gain at gate 1.0 is DetA: a returning player is re-attached immediately instead of spending frames as an unconfirmed new track (FN 13 026 → 623 for adaptive). On SoccerNet the decisive parameter is `reid_proximity_threshold`. The gate drops a pair when `1 − IoU` exceeds it, so 0.99 still requires `IoU ≥ 0.01` and only 1.0 disables it. A player who walks out of frame has a prediction extrapolated out of frame and zero overlap with the returning detection; the reference Deep OC-SORT recovers that with a geometry-only OCR pass against the last observed box, which BoT-SORT does not have, so here appearance at 1.0 is the only route. At 0.99 adaptive is +0.87 HOTA over tuned geometry (in-frame occlusions only); at 1.0 it is +1.23 with IDF1 +2.8, and `botsort` +2.29 with the switch count above. Both carry their closed-gate parameter sweeps to 1.0 unchanged. The same opening hurts DanceTrack and MOT17, where targets stay in frame: at 1.0 with their best DanceTrack settings adaptive loses 0.83 HOTA and `min` fusion loses 10 with seven times the ID switches. So it is documented as domain guidance, not a new default, with adaptive as the rule to prefer when the gate is open. Disabling the adaptive bonus (`cap=0`) changes HOTA by +0.12 / −0.13, so the gain is the additive form rather than the discriminativeness weighting. MOT17 is within single-sequence noise either way.

No post-processing interpolation; absolute numbers are not comparable to the paper's.

## Tests

`tests/core/test_botsort_reid.py`: adaptive fusion unit tests including the gate-before-gap regression (a gated competitor must not inflate the bonus), top-two gap with fewer than two candidates, proximity uses standard IoU under GIoU/DIoU/CIoU, negative-weight rejection, and end-to-end tracker tests for both fusions. Full suite: 1531 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/evaluations/results.md
tstanczyk95, line 14: Let's update "v.2.6.0.". Since now we're introducing  new parameters and values, e.g. this same file, lines 320-326.

src/trackers/core/botsort/tracker.py
```diff
@@ -95,11 +99,31 @@ class BoTSORTTracker(BaseTracker):
...
             is used. Computed from true IoU even when ``iou`` is GIoU/DIoU/CIoU.
             Default ``0.5`` (BoT-SORT ``proximity_thresh``; requires
             ``IoU >= 1 - reid_proximity_threshold``).
+        reid_fusion: How appearance is combined with geometry. ``"botsort"``
+            (default) takes the minimum of the geometry and appearance costs.
+            ``"adaptive"`` adds an adaptively weighted appearance term, which
+            weighs appearance more heavily when the best match stands clear of
+            the runner-up. The two produce different similarity ranges, so
+            ``minimum_iou_threshold_first_assoc`` needs retuning when switching
+            (see :func:`~trackers.core.reid.fusion.fuse_adaptive_reid_association`).
```
tstanczyk95, line 114: We should also mention about `minimum_iou_threshold_unconfirmed_assoc` there. Since it is also affected. Here in the same files, lines 380-396 (step 3).

docs/guides/reid.md
```diff
@@ -125,26 +129,131 @@ It is not the area where the shaded bands cross in the figure. That is two perce
...
+
+Two consequences are worth knowing before switching:
+
+- **The similarity range changes.** `"botsort"` returns values in `[0, 1]`; `"adaptive"` returns `[0, 1 + reid_appearance_weight + reid_adaptive_weight_cap]`, which is `[0, 2.25]` at the defaults. `minimum_iou_threshold_first_assoc` is applied to that fused value, so a threshold tuned for one method is a different gate under the other. Retune it when you switch, or the comparison measures the gate rather than the fusion.
```
tstanczyk95, line 170: Same remark as here: https://github.com/roboflow/trackers/pull/571/changes#r3972264147

We should also mention about `minimum_iou_threshold_unconfirmed_assoc` there. Since it is also affected. In `src/trackers/core/botsort/tracker.py`, lines 380-396 (step 3).

src/trackers/core/botsort/tracker.py
```diff
@@ -176,6 +204,18 @@ def __init__(
...
         self.reid_appearance_threshold = reid_appearance_threshold
         self.reid_proximity_threshold = reid_proximity_threshold
+        if reid_fusion not in get_args(ReidFusionMethod):
+            raise ValueError(f"Unknown reid_fusion {reid_fusion!r}. Valid options are: {get_args(ReidFusionMethod)}.")
+        if reid_appearance_weight < 0.0:
+            raise ValueError(f"reid_appearance_weight must be non-negative, got {reid_appearance_weight}")
+        if reid_adaptive_weight_cap < 0.0:
+            raise ValueError(f"reid_adaptive_weight_cap must be non-negative, got {reid_adaptive_weight_cap}")
+        if not 0.0 <= reid_appearance_floor <= 1.0:
+            raise ValueError(f"reid_appearance_floor must be in [0, 1], got {reid_appearance_floor}")
```
tstanczyk95, line 214: It would be good to also capture such cases as: `float("nan"), float("inf")`.

E.g.
`if not np.isfinite(reid_appearance_weight) or reid_appearance_weight < 0.0:`

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### 🚀 Added
 
+- **`BoTSORTTracker(reid_fusion="adaptive")`** — Deep OC-SORT's adaptive appearance weighting as an alternative to BoT-SORT's `min(d_iou, d_app)` fusion: adds a weighted appearance term whose weight grows when the best match stands clear of the runner-up. New `reid_appearance_weight` and `reid_adaptive_weight_cap` parameters; default fusion is unchanged. `reid_appearance_threshold` has no effect under adaptive fusion. `fuse_adaptive_reid_association` lives in `trackers.core.reid.fusion` beside the BoT-SORT rule so other trackers can reuse it ([#571](https://github.com/roboflow/trackers/pull/571)).
```
tstanczyk95, line 11: Let's mention `reid_appearance_floor` here as well. Especially that it's now included in `src/trackers/core/botsort/tracker.py`.

tests/core/test_botsort_reid.py
```diff
@@ -286,3 +290,225 @@ def test_real_reid_model_runs_over_frames(self) -> None:
...
+
+    def test_unknown_fusion_raises(self) -> None:
+        with pytest.raises(ValueError, match="Unknown reid_fusion"):
+            BoTSORTTracker(reid_fusion="deepocsort")  # type: ignore[arg-type]
+
+    @pytest.mark.parametrize("parameter", ["reid_appearance_weight", "reid_adaptive_weight_cap"])
+    def test_rejects_negative_adaptive_weights(self, parameter: str) -> None:
+        with pytest.raises(ValueError, match=parameter):
+            # mypy cannot narrow a dynamic dict against typed kwargs.
+            BoTSORTTracker(enable_cmc=False, **{parameter: -0.1})  # type: ignore[arg-type]
```
tstanczyk95, line 460: Let's also add test for the cases of `float("nan"), float("inf")` as mentioned here: https://github.com/roboflow/trackers/pull/571/changes#r3972306299

src/trackers/core/reid/fusion.py
```diff
@@ -65,3 +75,113 @@ def fuse_botsort_reid_association(
...
+    """Fuse IoU and appearance with Deep OC-SORT's adaptive weighting.
+
+    Adds a weighted appearance term to the geometric similarity instead of
+    taking a minimum. The weight grows when the best appearance match stands
+    clear of the runner-up and falls back to ``reid_appearance_weight`` when
+    the top candidates are hard to tell apart.
+
+    Implements equations (4)-(6) of the Deep OC-SORT paper, matching
+    ``compute_aw_new_metric`` in the authors' ``integrated_ocsort_embedding``
+    tracker, the variant behind their published results.
```
tstanczyk95, line 113: Let's write explicitly here that this is Deep OC-SORT's adaptive appearance weighting applied on top of BoT-SORT's score-fused geometry, rather than the full Deep OC-SORT association cost.

src/trackers/core/reid/fusion.py
```diff
@@ -65,3 +75,113 @@ def fuse_botsort_reid_association(
...
+        return np.zeros(similarity.shape[1 - axis], dtype=float)
+    partitioned = np.partition(similarity, -2, axis=axis)
+    best = np.take(partitioned, -1, axis=axis)
+    runner_up = np.take(partitioned, -2, axis=axis)
+    return np.minimum(best - runner_up, cap)
+
+
+def fuse_adaptive_reid_association(
+    association_similarity: np.ndarray,
+    appearance_similarity: np.ndarray,
```
tstanczyk95, line 96: How about adding shape validation?
```
association_similarity.shape == appearance_similarity.shape
association_similarity.shape == proximity_iou_similarity.shape
np.isfinite(...)
```

Especially that later we have:
```
appearance = np.clip(...)
...
out_of_range = ...
...
return association_similarity + ...
```

It might produce an error, but it might also produce some bad result and propagate it further otherwise, which might be hard to find later ;)

tests/core/test_botsort_reid.py
```diff
@@ -286,3 +290,225 @@ def test_real_reid_model_runs_over_frames(self) -> None:
         assert len(tracker.tracks) == 1
         bank = tracker.tracks[0].feature_bank
         assert bank is not None and bank.feature is not None
+
+
```
tstanczyk95, line 294: Maybe we can also add a test to assure that `reid_appearance_threshold` is ignored in adaptive mode?

Since it is directly mentioned in ‎`docs/guides/reid.md` (currently line 166).

It might be useful if we do even more modifications to BoT-SORT re-ID in the future.

tstanczyk95 (commented):
@AlexBodner very nice work and a cool feature to be added!

As we discussed earlier online, could you please provide the complete results so that for each of the four datasets: MOT17, SportsMOT, SoccerNet-tracking, DanceTrack, we have two sets of the results:

- BoT-SORT run with a re-ID not tuned on the dataset, e.g the one pre-trained on MSMT17
- BoT-SORT run with a re-ID tuned on the considered dataset, e.g. for the performance evaluation on SportsMOT, re-ID tuned on SportsMOT.

No need to worry about hyperparameter tuning for now, let's just focus on re-ID model at the moment, with a constant set of reasonable hyperparameter values.

So that we have complete information and see how much the re-ID is helping and how worth it is to tune it on the dataset of interest :)

Let's also add `reid_appearance_floor` in the PR description so that we have it for future references if needed ;)

tests/core/test_botsort_reid.py
AlexBodner: done

src/trackers/core/botsort/tracker.py
AlexBodner: done

src/trackers/core/botsort/tracker.py
AlexBodner: got it, done

src/trackers/core/reid/fusion.py
AlexBodner: not sure if doing so, because its something we handle ourselves internally and that is already verified to work separately. Would be redudant checking.

src/trackers/core/reid/fusion.py
AlexBodner: done

tests/core/test_botsort_reid.py
AlexBodner: done

CHANGELOG.md
AlexBodner: will add complete changelog when preparing release

docs/guides/reid.md
AlexBodner: done here as well

docs/evaluations/results.md
AlexBodner: done. I put provisionary version 2.7.0

PR #573, opened 2026-08-20
perf(mcbyte): skip mask-only work without mask evidence

JESUSROYETH:
McByte defaults to `enable_mask_manager=False`, so the mask-disabled path is the common one, not an edge case. Its three association stages still ran into mask-only candidate construction, though: `condition_similarity_with_masks()` copied and reduced the full similarity matrix, built the ambiguity and optional isolated-candidate matrices, and only after that checked whether there was usable mask evidence. On a default tracker those matrices got built for nothing, then discarded.

## The fix

The function keeps the same order as before: input validation, clear-match locking, reduced assignment. It now returns right there, before mask-only candidate construction starts, when no tracklet can actually receive mask evidence. That covers a few cases: the output or mask array is missing, the mask array has zero length, or the tracklet-to-mask or confidence mapping is empty or missing. With usable mask evidence, candidate construction and score updates stay the same as before.

## Measured

| benchmark | before | after | delta |
|---|---|---|---|
| microbench (µs/call, 1,091 matrices, MOT17-04 FRCNN, median [range], n=11) | 21.090 [20.832, 21.315] | 12.709 [12.603, 12.850] | -39.74% |
| DanceTrack e2e (ms, 25 seq / 25,508 frames / 225,148 boxes, median [range], 7 pairs) | 10,224.90 [10,046.24, 10,669.01] | 9,780.26 [9,600.19, 10,371.11] | -4.35% |
| MOT17-04 e2e (ms, median [range], 7 pairs) | 531.68 [511.18, 550.79] | 507.51 [499.67, 537.14] | -4.55% |

For each dataset five of seven pairs came out faster and two slower. The ranges overlap, so I'd call the end-to-end numbers supportive but machine-dependent, not a clean win.

## Correctness

5,000 random mask/no-mask cases give the exact same bit-level digest before and after, with zero mutations of the caller's matrix. I also compared a real MOT17-04 run, pairing 15,161 boxes by IoU (not by index) between the two implementations: minimum IoU 1.0, zero coordinate-bit or tracker-ID mismatches.

## Tests

A regression test covers the six no-evidence states, checks the candidate builders don't get called, and verifies both the locked/reduced result and that the caller's matrix stays untouched. Mask-association suite: 40 passed. Full non-integration suite: 1,554 passed, 3 skipped, 14 deselected. Pre-commit clean (ruff, formatting, codespell, mypy).

### DIFF
--- a/src/trackers/core/mcbyte/mask_association.py
+++ b/src/trackers/core/mcbyte/mask_association.py
@@ -231,9 +231,7 @@ def _apply_mask_similarity_boosts(
     remaining_detection_indices: list[int],
     tracklet_ids: list[int],
     detection_boxes: np.ndarray,
-    masks: np.ndarray,
-    tracklet_mask_dict: dict[int, int],
-    mask_avg_prob_dict: dict[int, float],
+    mask_output: MaskOutput,
     minimum_mask_average_confidence: float,
     minimum_mask_coverage: float,
     minimum_mask_fill_ratio: float,
@@ -253,7 +251,7 @@ def _apply_mask_similarity_boosts(
     For every candidate pair, the function:
 
         1. resolves the stable tracklet ID associated with the original row;
-        2. finds the corresponding local mask index in ``tracklet_mask_dict``;
+        2. finds the corresponding local mask index in ``mask_output``;
         3. verifies that the mask index and average mask confidence are valid;
         4. computes mask coverage and mask fill ratio for the original
            detection box;
@@ -284,10 +282,11 @@ def _apply_mask_similarity_boosts(
         detection_boxes: Detection boxes in ``xyxy`` format, ordered according
             to the columns of the original full association matrix. Expected
             shape is ``(num_detections, 4)``.
-        masks: Current propagated masks with shape ``(N, H, W)``.
-        tracklet_mask_dict: Mapping from stable tracklet IDs to local mask-array
-            indices.
-        mask_avg_prob_dict: Average mask confidence keyed by stable tracklet ID.
+        mask_output: Current propagated mask output. Its
+            ``tracklet_mask_dict`` maps stable tracklet IDs to local mask-array
+            indices, while ``mask_avg_prob_dict`` stores average mask confidence
+            keyed by stable tracklet ID. Callers must have already established
+            that masks and confidences are present.
         minimum_mask_average_confidence: Minimum average propagated-mask
             confidence required before mask evidence may be used.
         minimum_mask_coverage: Minimum fraction of the complete visible mask
@@ -298,6 +297,9 @@ def _apply_mask_similarity_boosts(
     Returns:
         None. ``conditioned_similarity`` is updated in place.
     """
+    if mask_output.masks is None or mask_output.mask_avg_prob_dict is None:
+        return
+
     # Local indices in the reduced matrix.
     candidate_rows, candidate_columns = np.where(candidate_matrix)
 
@@ -305,12 +307,12 @@ def _apply_mask_similarity_boosts(
     mask_areas: dict[int, int] = {}
     for local_track_index in np.unique(candidate_rows):
         tracklet_id = tracklet_ids[remaining_track_indices[local_track_index]]
-        mask_index = tracklet_mask_dict.get(tracklet_id)
-        if mask_index is None or not 0 <= mask_index < masks.shape[0]:
+        mask_index = mask_output.tracklet_mask_dict.get(tracklet_id)
+        if mask_index is None or not 0 <= mask_index < mask_output.masks.shape[0]:
             continue
 
         if mask_index not in mask_areas:
-            mask_areas[mask_index] = int(masks[mask_index].astype(bool, copy=False).sum())
+            mask_areas[mask_index] = int(mask_output.masks[mask_index].astype(bool, copy=False).sum())
 
     for local_track_index, local_detection_index in zip(
         candidate_rows,
@@ -321,19 +323,19 @@ def _apply_mask_similarity_boosts(
 
         # Resolve stable tracklet ID
         tracklet_id = tracklet_ids[original_track_index]
-        mask_index = tracklet_mask_dict.get(tracklet_id)
+        mask_index = mask_output.tracklet_mask_dict.get(tracklet_id)
         if mask_index is None:
             continue
 
-        if not 0 <= mask_index < masks.shape[0]:
+        if not 0 <= mask_index < mask_output.masks.shape[0]:
             continue
 
-        average_confidence = mask_avg_prob_dict.get(tracklet_id)
+        average_confidence = mask_output.mask_avg_prob_dict.get(tracklet_id)
         if average_confidence is None or average_confidence < minimum_mask_average_confidence:
             continue
 
         metrics = _get_mask_metrics_with_visible_area(
-            mask_bool=masks[mask_index].astype(bool, copy=False),
+            mask_bool=mask_output.masks[mask_index].astype(bool, copy=False),
             detection_xyxy=detection_boxes[original_detection_index],
             visible_mask_area=mask_areas[mask_index],
         )
@@ -475,13 +477,6 @@ def condition_similarity_with_masks(
         )
     ].copy()
 
-    conditioned_association = MaskConditionedAssociation(
-        conditioned_similarity=reduced_similarity,
-        locked_matches=locked_matches,
-        remaining_track_indices=remaining_track_indices,
-        remaining_detection_indices=remaining_detection_indices,
-    )
-
     if (
         mask_output is None
         or mask_output.masks is None
@@ -489,7 +484,13 @@ def condition_similarity_with_masks(
         or not mask_output.tracklet_mask_dict
         or not mask_output.mask_avg_prob_dict
     ):
-        return conditioned_association
+        # No tracklet can receive mask evidence, so the reduced problem is final.
+        return MaskConditionedAssociation(
+            conditioned_similarity=reduced_similarity,
+            locked_matches=locked_matches,
+            remaining_track_indices=remaining_track_indices,
+            remaining_detection_indices=remaining_detection_indices,
+        )
 
     # Ambiguity is a property of the original association situation before any
     # modifications, hence computed from base_similarity.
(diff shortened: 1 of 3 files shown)


Borda:
> CPU microbenchmark over 1,091 association matrices captured from real MOT17-04 FRCNN detections, 11 repetitions: 21.090 µs/call median `[20.832, 21.315]` down to 12.709 `[12.603, 12.850]`, -39.74%, disjoint ranges.
> 
> End-to-end, I ran seven alternating fresh-process pairs on each dataset. DanceTrack (25 sequences, 25,508 frames, 225,148 boxes) went from 10,224.90 ms median `[10,046.24, 10,669.01]` to 9,780.26 `[9,600.19, 10,371.11]`, -4.35%. MOT17-04 went from 531.68 ms `[511.18, 550.79]` to 507.51 `[499.67, 537.14]`, -4.55%. For each dataset five of seven pairs came out faster and two slower. The ranges overlap, so I'd call the end-to-end numbers supportive but machine-dependent, not a clean win.

This looks good; let's re-format it as a table and ensure that the HOMTA/MOTA metrics do not regress

JESUSROYETH:
mcbyte isn't in `ALL_TRACKER_IDS`, so there's no existing pinned-reference test for it here.. I ran the same method by hand: feed ground-truth boxes into `McByteTracker(enable_cmc=False, enable_mask_manager=False)`, write MOT files, run them through `evaluate_mot_sequences` (the repo's own TrackEval-parity evaluator) on DanceTrack (25 seq) and SportsMOT (45 seq).

| dataset | HOTA | MOTA | IDF1 | IDSW |
|---|---|---|---|---|
| DanceTrack (25 seq) | 81.21 | 99.63 | 78.31 | 587 |
| SportsMOT (45 seq) | 86.45 | 99.17 | 81.79 | 721 |

Ran it twice in the same tree, `mask_association.py` stashed back to `develop@35dd78b` for baseline, patch applied for the second run, nothing else changed. Every value above came out identical to the decimal. That's expected and not a coincidence, the Correctness section already showed the output digest is bit-identical over 5,000 cases with zero mismatches on the MOT17-04 IoU pairing, so any metric computed on that output has to come out the same. This just makes it explicit with the metrics instead of leaving it as an inference.

Updated the Measured section to a table too.

src/trackers/core/mcbyte/mask_association.py
```diff
@@ -513,15 +527,12 @@ def condition_similarity_with_masks(
         remaining_detection_indices=remaining_detection_indices,
         tracklet_ids=tracklet_ids,
         detection_boxes=detection_boxes,
-        mask_output=mask_output,
+        masks=mask_output.masks,
+        tracklet_mask_dict=mask_output.tracklet_mask_dict,
+        mask_avg_prob_dict=mask_output.mask_avg_prob_dict,
```
tstanczyk95: Why do you need to pass mask_output.masks, mask_output.tracklet_mask_dict and mask_output.mask_avg_prob_dict separately rather than mask_output as a whole and thus change the function signature?

src/trackers/core/mcbyte/mask_association.py
```diff
@@ -513,15 +527,12 @@ def condition_similarity_with_masks(
...
         minimum_mask_fill_ratio=minimum_mask_fill_ratio,
     )
 
-    return MaskConditionedAssociation(
-        conditioned_similarity=reduced_similarity,
-        locked_matches=locked_matches,
-        remaining_track_indices=remaining_track_indices,
-        remaining_detection_indices=remaining_detection_indices,
-    )
+    return conditioned_association
```
tstanczyk95: This relies on `conditioned_association.conditioned_similarity` and `reduced_similarity` sharing the same ndarray. Could we construct the result after applying the boosts, while handling the no-mask path separately, to make the mutation semantics more explicit?

Otherwise it might be confusing to another dev as reduced_similarity is modified inside `_apply_mask_similarity_boosts`.

src/trackers/core/mcbyte/mask_association.py
```diff
@@ -477,6 +475,22 @@ def condition_similarity_with_masks(
         )
     ].copy()
 
+    conditioned_association = MaskConditionedAssociation(
+        conditioned_similarity=reduced_similarity,
+        locked_matches=locked_matches,
+        remaining_track_indices=remaining_track_indices,
+        remaining_detection_indices=remaining_detection_indices,
+    )
```
tstanczyk95: See this remark of mine:
https://github.com/roboflow/trackers/pull/573/changes#r3846100215

tstanczyk95 (commented):
Hi @JESUSROYETH! 👋

Thank you for your valuable contribution! It overall looks very good!

I thoroguly checked your PR and I have some remarks. Here below and in the code. Could you reply to and/or resolve all of them? Thanks!

Remarks:

**1.**  
> the mask-disabled path is the common one, not an edge case

Just for clarity, this is the default setting as the heavy mask dependencies are not installed with pip install trackers (one needs to perform pip install trackers[mask] to install the mask support).

Normally though, McByte needs the mask for its desired behavior. Otherwise, it behaves similarly to BoT-SORT.

**2.**
> I ran the same method by hand: feed ground-truth boxes into McByteTracker(enable_cmc=False, **enable_mask_manager=False**)

So if I understand, you re-run McByte with your changes applied, but only without the mask enabled? Do you have the results with masks enabled? I mean HOTA, IDF1 and MOTA. We also need to see these before proceeding. Notably, the difference before and after your changes.

**3.**
> “I ran the same method by hand: feed **ground-truth boxes** into McByteTracker(enable_cmc=False, enable_mask_manager=False)

What do you mean by ground-truth boxes specifically? The track ground truth data  without track IDs? Or actual detections? Normally, we prefer the result comparison with detections, e.g. coming from YOLOX as used for our benchmarks on test splits. Detections are imperfect and thus can expose more performance differences when modifying the tracker's code, while ground truth bounding boxes are idealized and might not exposed as much.

**4.**

> Every value above came out identical to the decimal.

That's very good and well appreciated. Please keep doing it 🙂

src/trackers/core/mcbyte/mask_association.py
JESUSROYETH: @tstanczyk95 : 
Good question — it isn't strictly required. I split the fields because the early return narrows `masks` and `mask_avg_prob_dict`, but mypy cannot carry that narrowing into a helper receiving `MaskOutput`, so non-Optional parameters encoded that precondition in the signature. Worth noting `tracklet_mask_dict` isn't Optional, so it didn't need to be split at all.

That said, I agree that keeping the related mask state together gives the helper a cleaner interface. I'll switch it back to accepting `MaskOutput` and narrow the two Optional fields inside the helper. The performance improvement is unchanged, the no-mask path still returns before constructing the candidate matrices :+1:

src/trackers/core/mcbyte/mask_association.py
JESUSROYETH: @tstanczyk95 :

You're right, that was relying on `conditioned_similarity` and `reduced_similarity` being the same array, which is easy to miss, more so on a frozen dataclass that reads as immutable.

Restructured as you suggested: the no-mask path builds and returns its own result, and the mask path builds it after the boosts have been applied. The mutation now happens on a plain local array before anything wraps it.

Output is unchanged, I compared both versions over 5000 random cases covering the mask and no-mask paths and the reduced matrix is bit-identical, with the caller's matrix untouched in both. Mask association suite: 40 passed.

JESUSROYETH:
Hi @tstanczyk95 👋

Thank you very much for the review, it was really helpful. The points about how to frame the change and how to validate it are fair, and I will keep this pattern in mind for the next PRs.

On 1, you are right and my wording was loose. Saying "the mask-disabled path is the common one" is true about a plain `pip install trackers`, but written like that it reads as if running McByte without masks were the normal way to use it, and it is not. The masks are the point of McByte, without them it ends up close to BoT-SORT. What the patch skips is work on a path that cannot produce any mask evidence, that is not an argument for running it that way.

On 3, by ground-truth boxes I meant the annotation boxes fed in as if they were detections, with the track IDs dropped. Masks were off in that run too. Your objection is correct: that makes it an oracle-detection setup, and therefore less sensitive to association changes than imperfect detector output. It was not a strong enough validation for the concern you raised.

On 2, I ran it again the way you asked, with masks enabled and real detections, using the repo's own benchmark instead of a script of mine: `trackers benchmark mcbyte --dataset=[mot17] --device=cuda`, which sets `enable_mask_manager=True` and runs SAM + Cutie on GPU. Data is MOT17-val from `download_dataset`, 7 sequences and 2652 frames, with the public FRCNN detections that ship with it (30711 boxes), converted to the xyxy layout the CLI expects. Baseline is 35dd78b, the merge-base of this PR, and after is 55bc82f, so the diff between the two arms is only the two commits here. Both arms run from the same virtualenv with `PYTHONPATH` pointed at a different worktree, so the dependency versions are identical.

To check run-to-run stability, I ran the baseline twice. Both repetitions matched at full metric precision and produced the same seven SHA-256 hashes. The PR run then produced those same hashes.

| metric | baseline 35dd78b | PR 55bc82f | delta |
| :----- | ---------------: | ---------: | ----: |
| HOTA   |           50.732 |     50.732 | 0.000 |
| IDF1   |           57.421 |     57.421 | 0.000 |
| MOTA   |           47.705 |     47.705 | 0.000 |
| IDSW   |              167 |        167 |     0 |

All seven per-sequence results also matched, and the seven MOTChallenge output files were byte-identical. COMBINED HOTA is 0.5073214955745664 in both runs if you want the full precision.

<details>
<summary>Per-sequence, identical on both arms</summary>

| sequence  |   HOTA |   IDF1 |   MOTA | IDSW |
| :-------- | -----: | -----: | -----: | ---: |
| MOT17-02  | 35.945 | 39.020 | 30.273 |   24 |
| MOT17-04  | 54.958 | 61.676 | 48.792 |   16 |
| MOT17-05  | 45.754 | 54.895 | 52.011 |   16 |
| MOT17-09  | 52.443 | 59.474 | 51.650 |    7 |
| MOT17-10  | 51.372 | 59.523 | 54.398 |   61 |
| MOT17-11  | 49.768 | 57.002 | 56.210 |   26 |
| MOT17-13  | 58.782 | 71.583 | 61.027 |   17 |

</details>

One thing worth stating, since an all-equal table can also mean the mask path never ran: I instrumented `condition_similarity_with_masks` on MOT17-09 to count how often it was actually reached with mask evidence. Out of 536 calls, 530 had usable mask evidence and the other 6 exercised the no-evidence path. In 196 calls at least one boost landed, modifying 498 association scores in total. So both the mask-conditioned path and the new no-evidence shortcut were exercised.

I should be clear about one substitution: these are the public FRCNN detections the repo hosts for MOT17-val, not YOLOX. The YOLOX numbers you use are on the test splits through Codabench, which I cannot score locally, and I did not find YOLOX detection files for the val split in the repo. These FRCNN detections still exercise the concern you raised: imperfect detections with misses and false positives instead of idealised boxes, and anyone can reproduce it with one `download_dataset` call. If you have the YOLOX detection files you use for the benchmarks, I am happy to run the same comparison on them.

Also, the absolute values above are not comparable with the published MOT17 table, they use a different split and a different detector. They are only meant for the before/after comparison.

tstanczyk95 (approved):
Hi @JESUSROYETH,

Great, thank you for your detailed answer and explanations. It is well appreciated!

@Borda It looks good to me now, it can be merged.

Issue #594, opened 2026-09-09
eval: make the MOT ground-truth class sets user-defined, and add MOT20


AlexBodner:
## Description

MOT preprocessing follows TrackEval, but the ground-truth class sets are hardcoded to MOT17. TrackEval branches on the benchmark and we do not:

```python
# trackeval/datasets/mot_challenge_2d_box.py
distractor_class_names = ['person_on_vehicle', 'static_person', 'distractor', 'reflection']
if self.benchmark == 'MOT20':
    distractor_class_names.append('non_mot_vehicle')
```

```python
# src/trackers/io/mot.py:31
# TODO(MOT20): MOT20 additionally treats non_mot_vehicle (6) as a distractor;
# thread a benchmark parameter through _prepare_mot_sequence to support it.
_DISTRACTOR_CLASSES = (2, 7, 8, 12)
```

Ground-truth rows fall into three buckets. **Scored**: `pedestrian` (1), where a matched detection is a TP. **Distractor**: dropped before the metrics, no penalty. **Neither**: a detection on top of it counts as a false positive. MOT20 puts `non_mot_vehicle` (6) in the second bucket and we put it in the third, so on MOT20 a tracker is charged false positives that TrackEval forgives. On a synthetic four-row sequence, same tracker, same boxes:

```
distractor set (2, 7, 8, 12)      TP=2 FP=1 FN=0   MOTA=0.5000
distractor set (2, 6, 7, 8, 12)   TP=2 FP=0 FN=0   MOTA=1.0000
```

The set is reached from `evaluate_mot_sequence`, `evaluate_mot_sequences`, `trackers eval` and every `tune` trial. MOT17, SportsMOT, DanceTrack and SoccerNet are unaffected, since `(2, 7, 8, 12)` is already correct for all four.

The task is to remove the hardcoded configuration, so MOT20 becomes a table entry and any other convention can be supplied by a user without waiting for a release.

## Proposed interface

This is a first suggestion. If there is a cleaner way to shape it, we are open to proposals, so comment before you write code.

```python
# src/trackers/eval/mot_classes.py (new leaf module)

MOTClassPreset = Literal["mot17", "mot20"]

@dataclass(frozen=True)
class MOTClassConfig:
    scored_classes: tuple[int, ...] = (1,)
    distractor_classes: tuple[int, ...] = (2, 7, 8, 12)

MOT_CLASS_PRESETS: dict[MOTClassPreset, MOTClassConfig] = {
[...]

PR #597, opened 2026-09-10
feat: make MOT ground-truth classes configurable

omkar-334:
# MOT class configuration for preprocessing

## Description

Make MOT ground-truth class handling configurable during preprocessing. The default remains MOT17 behavior, while the MOT20 preset treats class 6 as a distractor. Custom scored and distractor class sets are also supported.

This change adds the shared class configuration and resolver, threads the resolved config through MOT preprocessing, exports the new evaluation types, and covers the default, MOT20, and custom-config cases with in-memory tests.

The evaluation API, CLI option, MOT20 warning, and external TrackEval/count validation are intentionally left for follow-up work.

## Validation

Static compilation and `git diff --check` pass. The focused MOT suite passes with 14 tests.


omkar-334:
Part of #594. This PR addresses items 1, 2, and 6:
  - Adds configurable MOT class presets and resolver
  - Threads class configuration through MOT preprocessing
  - Adds tests covering default, MOT20, and custom class configurations

